PR #18, opened 2024-07-26
Add Linters for PHP and CSS

### LINKED ISSUE #5: Add linters for PHP and CSS
The theme needs to follow the [WordPress coding standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/) and tools like [PHPCS](https://github.com/WordPress/WordPress-Coding-Standards) and stylelint can be added to the project to help maintain the code quality.

(It is less likely that the theme will need custom JavaScript)

amjadr360:
Ref https://github.com/WordPress/twentytwentyfive/issues/5

~~"Copy & paste from twentytwentyfour."~~  

Initial setup copied from Twenty Twenty-Four. Updated dependencies and made further improvements based on feedback.

## Testing Instructions

To run the tests locally, you'll need:

- **Node.js:** It's recommended to install Node.js using `nvm`, which helps manage the Node.js version required by the project.
- **Composer:**

Once you have these tools set up, install the necessary development dependencies by running:

```bash
npm i && composer install
```

You can then use the following commands to test the code:

- **`npm run lint:css`**: Lints all CSS files.
- **`npm run lint:css:fix`**: Lints and attempts to autofix any issues in the CSS files.
- **`npm run lint:php`**: Checks PHP files for syntax and standards errors according to [WordPress coding standards](https://developer.wordpress.org/coding-standards/).
- **`npm run lint:php:fix`**: Attempts to automatically fix PHP errors.


carolinan:
Hi
I am not able to review the GitHub actions because I don't know anything about it :)
but I would like to make the following change requests:

The downside with copying the setup from the Twenty Twenty-Four repository is that it is already out of date.

- In my **opinion** it would be best if the theme supported the same versions of node as Core and Gutenberg, instead of 16 which is not supported anymore (EOL).
```
"node": ">=20.10.0",
"npm": ">=10.2.3"
```
- I assume that we want to use WordPressCS 3.1.0, not 2.2? Or are there any reasons like known conflicts that prevents this?
- The latest wp stylelint config is 22.5.0.

luminuu:
I agree, the versions should be updated to the latest used by Gutenberg/Core.

amjadr360:
Thank you for looking into the PR and providing feedback. I’ve updated the dependencies to the latest versions.

composer.json
```diff
@@ -0,0 +1,30 @@
+{
+	"name": "wordpress/twentytwentyfive",
```
juanfra, line 2: Should we include the PHP min version in this file?

juanfra (commented):
Thank you for working on this! Got it working locally. 

We'll also need to add a **Testing** section in the `readme` file, so that contributors are familiar with how to run these.

composer.json
amjadr360: I think we should, since the theme’s minimum requirement is PHP 7.2.24+. Should we add PHP 7.2.24+?

README.md
```diff
@@ -42,6 +42,26 @@ To get started with development:
 
 Some theme features / PRs may require Gutenberg trunk and will be described or tagged accordingly.
 
+## Testing
+
+To run the tests locally, you'll need:
+
+- **Node.js:** It's recommended to install Node.js using `nvm`, which helps manage the Node.js version required by the project.
```
vcanales: Let's add a link to nvm installation instructions:

```suggestion
- **Node.js:** It's recommended to install Node.js using `nvm`, which helps manage the Node.js version required by the project. Installation instructions [here](https://github.com/nvm-sh/nvm?tab=readme-ov-file#installing-and-updating).
```

README.md
```diff
@@ -42,6 +42,26 @@ To get started with development:
...
+- **Composer:**
```
vcanales:
```suggestion
- **Composer:** you can find the instructions for your OS [here](https://getcomposer.org/doc/00-intro.md#system-requirements).
```

vcanales (approved):
Looks good to me! I've left a couple of suggestions for expanding on where to get the dependencies for local testing.

juanfra (approved):
Thank you so much for working on this Ajmad! and thanks Vicente for the review.

PR #21, opened 2024-08-14
Add a block binding and pattern for the copyright date

### LINKED ISSUE #7: Add block bindings for copyright date and privacy policy link
### What
Add a block binding for a dynamic copyright date.
Add a block binding for a link to the privacy policy page. This link should only display if the privacy policy page is published.

### Why
The purpose of this task is twofold: To solve two common problems and showcase how to use the block bindings api.

### Resources
https://make.wordpress.org/core/2024/03/06/new-feature-the-block-bindings-api/
https://developer.wordpress.org/news/2024/02/20/introducing-block-bindings-part-1-connecting-custom-fields/
https://developer.wordpress.org/news/2024/03/06/introducing-block-bindings-part-2-working-with-custom-binding-sources/
https://developer.wordpress.org/news/2024/05/06/building-a-book-review-site-with-block-bindings-part-1-custom-fields-and-block-variations/
https://developer.wordpress.org/news/2024/06/06/building-a-book-review-site-with-block-bindings-part-2-queries-patterns-and-templates/

carolinan:
This PR registers a block binding source with a custom function as a callback.
The callback returns a text with the copyright symbol and the year. The purpose is for the year to update automatically without the user needing to edit the pattern.

The binding is then used in a block pattern with a paragraph block.

Closes https://github.com/WordPress/twentytwentyfive/issues/7

Note that the development version of the Gutenberg plugin was active when these screenshots were taken:

Editor:
image

Front:
image

And here it is in the Portfolio pattern, above the footer.
image


functions.php
```diff
@@ -7,3 +7,29 @@
...
+function twentytwentyfive_copyright_binding() {
+	/* translators: Copyright date format, see https://www.php.net/manual/datetime.format.php */
+	$date_format = _x( 'Y', 'copyright date format', 'twentytwentyfive' );
+	/* translators: 1: Copyright symbol or word, 2: Date, 3: Site title */
+	$copyright_text = sprintf(
+		__( '%1$s %2$s %3$s', 'twentytwentyfive' ),
+		'&copy;',
+		wp_date( $date_format ),
+		get_bloginfo( 'name' )
+	);
```
g-elwell: please move this translators comment to the line immediately preceding the i18n function call, otherwise it won't be detected
```suggestion
	$copyright_text = sprintf(
		/* translators: 1: Copyright symbol or word, 2: Date, 3: Site title */
		__( '%1$s %2$s %3$s', 'twentytwentyfive' ),
		'&copy;',
		wp_date( $date_format ),
		get_bloginfo( 'name' )
	);
```

carolinan: Thank you.

functions.php
```diff
@@ -133,3 +133,44 @@ function twentytwentyfive_pattern_categories() {
...
+	 *
+	 * @since Twenty Twenty-Five 1.0
+	 * @return string Copyright text.
+	 */
+	function twentytwentyfive_copyright_binding() {
+		/* translators: Copyright date format, see https://www.php.net/manual/datetime.format.php */
+		$date_format    = _x( 'Y', 'copyright date format', 'twentytwentyfive' );
+		$copyright_text = sprintf(
+			/* translators: 1: Copyright symbol or word, 2: Date */
+			__( '%1$s %2$s', 'twentytwentyfive' ),
```
juanfra: I believe we may want to use an escaping function here, as `wp_date` is hookable, and it could lead into some security issues.

functions.php
```diff
@@ -133,3 +133,44 @@ function twentytwentyfive_pattern_categories() {
...
+	}
+	/**
+	 * Callback function for the copyright block binding source.
+	 *
+	 * @since Twenty Twenty-Five 1.0
+	 * @return string Copyright text.
+	 */
+	function twentytwentyfive_copyright_binding() {
+		/* translators: Copyright date format, see https://www.php.net/manual/datetime.format.php */
+		$date_format    = _x( 'Y', 'copyright date format', 'twentytwentyfive' );
```
juanfra: I'm not sure about passing this one through a translation function. 

If it were something other than the year, and it needed a date format it would make sense. I can't think of any language with a different format than the four numbers.

functions.php
```diff
@@ -133,3 +133,44 @@ function twentytwentyfive_pattern_categories() {
 endif;
 
 add_action( 'init', 'twentytwentyfive_pattern_categories' );
+
+add_action( 'init', 'twentytwentyfive_register_block_bindings', 9 );
+
+/**
+ * Register block bindings.
+ */
+if ( ! function_exists( 'twentytwentyfive_register_block_bindings' ) ) :
```
juanfra, line 140: Maybe we wrap each function in its own `if ( ! function_exists(...))`?

functions.php
```diff
@@ -133,3 +133,44 @@ function twentytwentyfive_pattern_categories() {
 endif;
 
 add_action( 'init', 'twentytwentyfive_pattern_categories' );
+
+add_action( 'init', 'twentytwentyfive_register_block_bindings', 9 );
```
juanfra: Is there any particular reason why you are setting the priority to 9?

We can move this one below the function to match what we did on the rest.

juanfra (commented):
I love this @carolinan 🏅

It's super exciting to be showcasing how to use bindings in the default theme. I left a few comments we may want to look at.

functions.php
carolinan: Well, it is intended to run before _something_, but I can't for the life of me remember what. I have removed it.

carolinan:
I think the placeholder text that shows in the editor can be improved, but I am reluctant to include the year since the date will not update in the editor once the block is placed.

The current text in the editor is `Copyright`.  
The current text on the front is `© 2024`

Because the block is synced, it is purple in the editor, and there is a block binding specific panel in the block settings sidebar for the attribute: Other than that it is difficult to see that this block is a placeholder and not just any other editable block.
The reason why I did not go with  `Copyright date placeholder` was that the text is so much longer than what is intended in the design, so the editor and front does not match well.

Perhaps the editor text should be `© YEAR` ? Does anyone have any other ideas?

juanfra:
I think the `© YEAR` proposal is the one that better reflects the pattern's functionality.

PR #28, opened 2024-08-15
Add Props Bot Action

### LINKED ISSUE #27: Use Props Bot Action
The [WordPress Props bot Action](https://github.com/WordPress/props-bot-action) is designed to ensure a consistent contributor recognition.  I think this project should use it as well.

aaronjorbin:
Fixes #27.

**Description**

Adds the Props Bot Action to help ensure contributors are properly recognized. Very demure. Very mindful.

**Screenshots**

N/A

**Testing Instructions**

This version is configured the same as what is used in the [Gutenberg repo](https://github.com/WordPress/gutenberg/blob/trunk/.github/workflows/props-bot.yml)

**Contributors**

This should make that easier to track :)

### DIFF
--- a/.github/workflows/props-bot.yml
+++ b/.github/workflows/props-bot.yml
@@ -18,7 +18,7 @@ on:
     # You cannot filter this event for PR comments only.
     # However, the logic below does short-circuit the workflow for issues.
     issue_comment:
-        type:
+        types:
             - created
     # This event will run everytime a new PR review is initially submitted.
     pull_request_review:
--- a/.github/workflows/props-bot.yml
+++ b/.github/workflows/props-bot.yml
@@ -1,0 +1,88 @@
+name: Props Bot
+
+on:
+    # This event runs anytime a PR is (re)opened, updated, marked ready for review, or labeled.
+    # GitHub does not allow filtering the `labeled` event by a specific label.
+    # However, the logic below will short-circuit the workflow when the `props-bot` label is not the one being added.
+    # Note: The pull_request_target event is used instead of pull_request because this workflow needs permission to comment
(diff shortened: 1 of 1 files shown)


.github/workflows/props-bot.yml
```diff
@@ -0,0 +1,88 @@
...
+            - opened
+            - synchronize
+            - reopened
+            - labeled
+            - ready_for_review
+    # This event runs anytime a comment is added or deleted.
+    # You cannot filter this event for PR comments only.
+    # However, the logic below does short-circuit the workflow for issues.
+    issue_comment:
+        type:
```
jeherve: You'll most likely want to update this to `types`. `type` is not supported:
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#issue_comment

```suggestion
        types:
```

jeherve (commented):
I wanted to suggest the same thing. 👍 I only have a minor suggestion, below.

.github/workflows/props-bot.yml
aaronjorbin: Thanks for the catch. Now to raise the fix over at https://github.com/WordPress/gutenberg/blob/trunk/.github/workflows/props-bot.yml#L21

jeherve: Since the problem is the same in the Gutenberg repo and the WordPress-develop repos, I opened PRs to fix the issue there as well:

https://github.com/WordPress/gutenberg/pull/64557
https://github.com/WordPress/wordpress-develop/pull/7203

jeherve (approved):

PR #32, opened 2024-08-16
Added new contributors

kafleg:
Added new contributors in the list.

### DIFF
--- a/CONTRIBUTORS.MD
+++ b/CONTRIBUTORS.MD
@@ -7,3 +7,6 @@
 | @MaggieCabrera | @onemaggie |
 | @kafleg | @kafleg |
 | @SeanAverS | |
+| @jeherve | @jeherve |
+| @aaronjorbin | @jorbin |
+| @juanfra | @juanfra |


carolinan:
I don't think this file needs to be maintained anymore because the props bot is used?
 @aaronjorbin

kafleg:
> I don't think this file needs to be maintained anymore because the props bot is used? @aaronjorbin

If that is the case, that is awesome.

aaronjorbin:
@desrosj should be able to confirm next week, but I believe you are correct that it won't be necessary to manually curate this.  

I think he has a more complex script for pulling contributors but a one liner like this can get a uniq list of every WordPress.org username who has co-authored.

```git log origin/trunk --format="%(trailers:key=Co-authored-by)" | sort | cut -d '<' -f 2 | grep 'git.wordpress' | cut -d '@' -f 1 | sort | uniq```

carolinan:
@markallasread If you would like credits (props), can you please connect your GitHub account to your WordPress.org account or add your usernames to this list.

desrosj:
The statements above are correct! As long as the `Co-authored-by` trailers are copied into the merge commit message, the props for all changes can be parsed out of the `git log`.

I recommend that anyone working with the repository check out the relevant section on the [Contributor Attribution page of the Core handbook](https://make.wordpress.org/core/handbook/best-practices/contributor-attribution-props/#github-repositories-merge-commits). Maybe adding this to the `README.md` file could help call attention to it?

PR #35, opened 2024-08-16
Added fonts

kafleg:
Added fonts listed in the styles of Figma.


carolinan:
Please organize the files in folders, one folder per font family. `assets/fonts/name`

It looks like the files may have been downloaded using the font library?
This makes it more difficult to identify which file is for which family, since the name is randomized. Long term, this can make it more difficult to maintain the fonts.

Please add the source and license information to the readme.txt file.

kafleg:
> Please organize the files in folders, one folder per font family. `assets/fonts/name`
> 
> It looks like the files may have been downloaded using the font library? This makes it more difficult to identify which file is for which family, since the name is randomized. Long term, this can make it more difficult to maintain the fonts.
> 
> Please add the source and license information to the readme.txt file.

Do I need to upload those fonts in woff2 format or ttf format is also fine?

carolinan:
Please use woff2. If it is not included in the google fonts zip, you should be able to get them here https://gwfh.mranftl.com/

juanfra:
Thanks for working on this! Adding to what Carolina said, please be sure that we keep the spacing with tabs instead of spaces for the `theme.json` file.

aaronjorbin:
`theme.json` wasn't updated after the files were renamed and put into folders

carolinan:
If any of these fonts come with variable fonts, those should be used (the google fonts page should say if it is a variable font or not and they would be included in the zip file)

readme.txt
```diff
@@ -33,5 +33,38 @@ This theme bundles the following third-party resources:
...
+License: SIL Open Font License, 1.1, https://opensource.org/licenses/OFL-1.1
+Source: https://github.com/mozilla/Fira
+
+Platypi Font
+License: SIL Open Font License, 1.1, https://opensource.org/licenses/OFL-1.1
+Source: https://github.com/d-sargent/platypi
+
+Ysabeau Office Font
+License: SIL Open Font License, 1.1, https://opensource.org/licenses/OFL-1.1
+Source: https://github.com/CatharsisFonts/Ysabeau
```
carolinan: Were the fonts downloaded from the GitHub repositories?
I am asking because sometimes the version on GitHub is not the same as on Google Fonts. Sometimes Google Fonts are not up to date.

kafleg: I downloaded this font from https://fonts.google.com/ 

So, do you want me to replace the font from GitHub?

carolinan: No we only need to do that if the Google font version is missing glyphs / have bugs. 

The URL can be kept in the readme, but lets not call it "source" if it is not the source, because it can get confusing when trying to update the fonts later.

So I think the file can include both the GitHub link for reference, and the google fonts link as the source?

kafleg: Thank you, that sounds good. I think we can do in this way. 

Example,

```
Fira Code Font
Copyright (c) 2014, The Fira Code Project Authors
License: SIL Open Font License, 1.1, https://opensource.org/licenses/OFL-1.1
Reference: https://github.com/tonsky/FiraCode 
Source: https://fonts.google.com/specimen/Fira+Code

```

Can you verify if this is the right way to do?

carolinan:
> If any of these fonts come with variable fonts, those should be used (the google fonts page should say if it is a variable font or not and they would be included in the zip file)

For example, Fira Code has a variable font inside the zip folder from Google fonts: FiraCode-VariableFont_wght, but it is not a woff2 format.
So it needs to be converted. What converter was used for the other true type files?

All fonts needs to be checked, I just checked one.

carolinan:
> > If any of these fonts come with variable fonts, those should be used (the google fonts page should say if it is a variable font or not and they would be included in the zip file)
> 
> For example, Fira Code has a variable font inside the zip folder from Google fonts: FiraCode-VariableFont_wght, but it is not a woff2 format. So it needs to be converted. What converter was used for the other true type files?
> 
> All fonts needs to be checked, I just checked one.

Do you want to make this update in this PR, or merge this and do follow ups?
Also in theme.json, the order of the font weights varies depending on the font, can we make this consistent? (This becomes irrelevant if the variable fonts are used, so that should be changed first)

kafleg:
> > > If any of these fonts come with variable fonts, those should be used (the google fonts page should say if it is a variable font or not and they would be included in the zip file)
> > 
> > 
> > For example, Fira Code has a variable font inside the zip folder from Google fonts: FiraCode-VariableFont_wght, but it is not a woff2 format. So it needs to be converted. What converter was used for the other true type files?
> > All fonts needs to be checked, I just checked one.
> 
> Do you want to make this update in this PR, or merge this and do follow ups? Also in theme.json, the order of the font weights varies depending on the font, can we make this consistent? (This becomes irrelevant if the variable fonts are used, so that should be changed first)

Can we merge this and I'll followup in the new PR. I will make the order of the fonts in ascending order.

Issue #37, opened 2024-08-16
Add headers


carolinan:
Figma link:
https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3056-195&t=qClGa6ofM1L6v4bG-4

I would recommend starting with two template parts: One for the horizontal header and one for the sidebar.
Each header can then be separate patterns, with the default headers placed in the respective template part.

- [x] Header (Main): Site title and navigation
- [x] Header: Site title and hamburger menu
- [x] Header: Columns with site title, tagline and vertical nav
- [x] Header: Big type

Sidebars
- [x] Header as sidebar 
- [ ] Header as sidebar with navigation

carolinan:
> Header: Site title and hamburger menu

Core already provides a header pattern called "centered header with logo", Should we use that instead?
https://wordpress.org/patterns/pattern/centered-header-with-logo/
Its markup uses spacing preset 30, 40, and the link color with the slug "contrast"

```
<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|40","right":"var:preset|spacing|30","bottom":"var:preset|spacing|40","left":"var:preset|spacing|30"}}},"layout":{"type":"constrained"}} -->
<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--40);padding-right:var(--wp--preset--spacing--30);padding-bottom:var(--wp--preset--spacing--40);padding-left:var(--wp--preset--spacing--30)"><!-- wp:group {"align":"wide","layout":{"type":"flex","justifyContent":"center","orientation":"vertical"}} -->
<div class="wp-block-group alignwide"><!-- wp:site-logo /-->

<!-- wp:site-title {"style":{"elements":{"link":{"color":{"text":"var:preset|color|contrast"}}}},"textColor":"contrast","fontSize":"large"} /-->

<!-- wp:navigation {"layout":{"type":"flex","setCascadingProperties":true,"justifyContent":"center"}} /--></div>
<!-- /wp:group --></div>
<!-- /wp:group -->

```

juanfra:
A PR with the four headers that are not "sidebar" has been merged [here](https://github.com/WordPress/twentytwentyfive/pull/52).

carolinan:
For now, the plan is to not include Header as sidebar with navigation, so I am closing this as complete.

PR #52, opened 2024-08-18
Add headers

carolinan:
**Description**
Add four header patterns. This PR does not include the headers that are sidebars.
Partial for https://github.com/WordPress/twentytwentyfive/issues/37

**Testing Instructions**
Activate Gutenberg. Note that I am testing with Gutenberg trunk.
If you don't want to test with Gutenberg, then just switch the header patterns manually by inserting them, since you wont see the "Design" panel.

Since the font sizes and font family is not in yet, the main thing to test in this PR is that:
- The blocks are the same as [in the designs.](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3056-195&t=fWImLHb7Ke2yTwcj-0)
- The spacing to the top, bottom, left and right of the header content matches the designs in Figma closely.

Apply the PR.
Confirm that the default header is used and displays correctly in the editor and front.
Test with different screen widths.

Then select the different heading patterns from the Design panel in the sidebar in the Site Editor, and repeat the test.
There is no need to test the patterns that are added by core. The patterns from the theme should be the four first patterns, at the top of the Design panel.


theme.json
```diff
@@ -110,16 +110,46 @@
...
 			"area": "header",
-			"name": "Header"
+			"name": "header",
+			"title": "Header"
 		},
 		{
 			"area": "footer",
-			"name": "Footer"
+			"name": "footer",
+			"title": "Footer"
```
carolinan, line 152: This is the same change as in https://github.com/WordPress/twentytwentyfive/pull/55, I wasn't sure which would get merged first.

patterns/header.php
```diff
@@ -0,0 +1,19 @@
...
+ * Categories: header
+ * Block Types: core/template-part/header
+ * Description: Header with site title and navigation
+ */
+?>
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
+<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--30);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--30);padding-left:var(--wp--preset--spacing--50)">
+	<!-- wp:site-title {"level":0} /-->
+	<!-- wp:group {"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group">
```
juanfra: We may want to set some spacing here. If not, there's a gap at the right of the navigation, which makes it look like it's not aligned with the dimensions of the main site wrapper.

Screenshot 2024-08-20 at 12 19 38

theme.json
```diff
@@ -110,16 +110,46 @@
 				"left": "var(--wp--preset--spacing--50)",
 				"right": "var(--wp--preset--spacing--50)"
 			}
+		},
+		"blocks":{
+			"core/site-title": {
+				"typography": {
+					"fontSize": "var:preset|fontSize|medium",
+					"fontWeight": "700",
+					"letterSpacing": "-2"
```
juanfra, line 119: I see in the figma that for some variations we go with `-2` and for some others `0`. We may want to revisit this in the future.

juanfra (commented):
Thanks so much for working on this, Carolina! It looks fantastic. I left a comment about one tiny thing I found with spacing for the default header.

patterns/header.php
carolinan: This group is only added to support adding a block hook after the navigation block. Yes it should probably have 0 padding set.

carolinan: Removed, thank you for catching this issue.

juanfra (approved):
Thanks, Carolina!

✅ LGTM

Issue #72, opened 2024-08-19
Customize TT5's link styles


beafialho:
Because the link styles usually look a bit rough, I suggest we add something like the following, unless someone has a better alternative, to improve the appearance of the theme's link underlines.

```
/*
 * Link styles
 * https://github.com/WordPress/gutenberg/issues/42319
 */
a {
	text-decoration-thickness: 1px !important;
	text-underline-offset: .1em;
}
```

Here's a before/after:
![Captura de ecrã 2024-08-19, às 11 35 51](https://github.com/user-attachments/assets/feb16c62-9ae2-4e4d-88d1-cc75c7286e2f)
![Captura de ecrã 2024-08-19, às 11 36 04](https://github.com/user-attachments/assets/3a918009-a8a7-413a-87ab-949d6225e379)

beafialho:
cc: @carolinan @juanfra @richtabor

carolinan:
I think we should add this to style.css (!) and not theme.json, for the same reasons mentioned in
https://github.com/WordPress/twentytwentyfive/issues/6

beafialho:
Yes, my intent was to add it to `style.css`

carolinan:
I wonder if there should also be an animation to ease it in and out?

beafialho:
> I wonder if there should also be an animation to ease it in and out?

That would be so cool!

melchoyce:
It's not currently possible to add an animation between `text-decoration: none` and `text-decoration: underline.` The only properties that can animate are `text-decoration-color` and `text-decoration-thickness.` I've tried a few iterations of animating `text-decoration-color` and don't think it'll work based on how the styles are set up. In order to add an animation, we'd need to refactor the underline styles to use pseudo-classes instead of `text-decoration` for link styles.

carolinan:
Thank you @melchoyce. I will close this issue as completed. 😞

Issue #6, opened 2024-07-03
Solving the focus outline style


carolinan:
This issue is inherited from Twenty Twenty-Four: https://github.com/WordPress/twentytwentyfour/issues/744
To comply with [WCAG 2.4.13 focus appearance](https://www.w3.org/TR/WCAG22/#focus-appearance), the focus indicator needs to be at least 2px thick and have a contrast ratio of at least 3:1 between the same pixels in the focused and unfocused states. See [Understanding success criterion](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance).
The open issue also suggests that the outline needs to be solid.

The editor lacks options for adjusting the outline for all focusable elements, so this needs to be solved with custom CSS.
To clarify, it is not enough to only style the focus outline for links and buttons.
The options are to add the CSS to theme.json or enqueue a stylesheet.
In the discussions about how to solve this for Twenty Twenty-Four, it was brought up that if the style is added to theme.json, users can remove it from the CSS field in the Site Editor, which would make the theme less accessible (Unless the user also adds an improved style).

Related Gutenberg issue: https://github.com/WordPress/gutenberg/issues/38277

richtabor:
> To clarify, it is not enough to only style the focus outline for links and buttons.

Can you clarify this?

carolinan:
The focus style needs to be accessible on **all** focusable elements.
The WCAG does not limit the guideline to links and buttons, it applies to all focusable elements.

joedolson:
E.g., `input`, `textarea`, `select`, `input`, `button`, `tabindex=!-1`, `contentEditable`, etc.

carolinan:
When I was working on the back to top block, I found that it was not reasonable for _me_ to create a complete list of focusable elements, I ended up with: 
https://github.com/WordPress/gutenberg/pull/50600/files#diff-7d8a7770b0e394828f51ea87f4de55be62d6e2f91a3a3951380ae8ebc4f15327R16

That is why in TT4, the style right now has the selector `where(.wp-site-blocks *:focus)`

This causes some styles for the navigation block that users have reported are unwanted, and is still discussed in
https://core.trac.wordpress.org/ticket/60808#comment:44

germanfrelo:
This is my suggestion:

```css
/**
 * Better focus indicators.
 * Source: https://www.sarasoueidan.com/blog/focus-indicators/#a-%E2%80%98universal%E2%80%99-focus-indicator
 */

[...]

PR #94, opened 2024-08-20
Try link and focus styles

carolinan:
**Description**
This PR adds text decoration thickness and underline offset for links. 
This style is enqueued both in the editors and the front, using style.css and editor-style.css

It also adds a solid outline with 2px width on focus, to all focusable elements.
This style is only enqueued on the front, using style.css.
I did not add this style the editor, because it affected the way the blocks look when they are selected, and _in my opinion_, this can make the blocks more difficult to work with.

The PR Uses `get_parent_theme_file_uri()` so that the files are enqueued also for child themes. Please correct me if this is wrong!

For https://github.com/WordPress/twentytwentyfive/issues/72 https://github.com/WordPress/twentytwentyfive/issues/6

**Screenshots**

**Testing Instructions**


functions.php
```diff
@@ -7,3 +7,24 @@
  * @package Twenty Twenty-Five
  * @since Twenty Twenty-Five 1.0
  */
+
+/**
+ * Enqueue stylesheet.css in the editors.
```
carolinan:
```suggestion
 * Enqueue editor-style.css in the editors.
```

functions.php
```diff
@@ -7,3 +7,24 @@
  * @package Twenty Twenty-Five
  * @since Twenty Twenty-Five 1.0
  */
+
+/**
+ * Enqueue editor-style.css in the editors.
+ */
+function twenty_twenty_five_editor_style() {
```
juanfra: To be consistent with the previous default themes we should use the theme prefix without underscores.
```suggestion
function twentytwentyfive_editor_style() {
```

functions.php
```diff
@@ -7,3 +7,24 @@
  * @package Twenty Twenty-Five
  * @since Twenty Twenty-Five 1.0
  */
+
+/**
+ * Enqueue editor-style.css in the editors.
```
juanfra, line 13: We need to document the `@since` and `@return` for the PHP functions. As well as the params, if any.

beafialho:
Could we add the hover state as well? Or maybe address that in another issue? I realised I haven't specified that in the Figma file.

beafialho:
I have added the links hover states [here](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=2431-1112&t=5PGQipVqkNPOWBux-4), if it's helpful. Let me know if I missed any.

carolinan:
If the hover styles can be added in theme.json, that would be better. If it doesn't work, we can use the css files.

functions.php
```diff
@@ -7,3 +7,24 @@
...
+ */
+function twenty_twenty_five_editor_style() {
+	add_editor_style( get_parent_theme_file_uri( 'editor-style.css' ) );
```
juanfra: In the codex page for [add_editor_style()](https://developer.wordpress.org/reference/functions/add_editor_style/) I see the following:

> If a child theme is used, both the current child and parent theme directories are tested and both the files with the same relative path are linked with this single call if they are found.

It was my understanding that it should work like that. But after testing it out, I can only get it working with the tweak you added. Maybe there's something that changed and it's not documented? Or maybe just a bug.

The other thing I was thinking of, is if we want to move this one out to the assets folder to keep things more organized.

functions.php
```diff
@@ -7,3 +7,24 @@
...
+}
+add_action( 'after_setup_theme', 'twenty_twenty_five_editor_style' );
```
juanfra:
```suggestion
add_action( 'after_setup_theme', 'twentytwentyfive_editor_style' );
```

functions.php
```diff
@@ -7,3 +7,24 @@
...
+
+/**
+ * Enqueue stylesheet.css on the front.
```
juanfra, line 24: Same comment about documenting the `@since` and `@return`.

functions.php
```diff
@@ -7,3 +7,24 @@
...
+ */
+function twenty_twenty_five_enqueue_styles() {
```
juanfra:
```suggestion
function twentytwentyfive_enqueue_styles() {
```

functions.php
```diff
@@ -7,3 +7,24 @@
...
+	wp_enqueue_style(
+		'twenty-twenty-five-style',
```
juanfra:
```suggestion
		'twentytwentyfive-style',
```

functions.php
```diff
@@ -7,3 +7,24 @@
...
+		get_parent_theme_file_uri( 'style.css' ),
```
juanfra, line 32: I believe `get_stylesheet_uri()` should be good here.

functions.php
```diff
@@ -7,3 +7,24 @@
...
+		array(),
+		wp_get_theme()->get( 'Version' ),
+	);
+}
+add_action( 'wp_enqueue_scripts', 'twenty_twenty_five_enqueue_styles' );
```
juanfra:
```suggestion
add_action( 'wp_enqueue_scripts', 'twentytwentyfive_enqueue_styles' );
```

juanfra (commented):
Thanks for working on this! I left a few comments.

functions.php
carolinan: `get_stylesheet_uri()` gets the active themes style.css. So it depends on how and if we want the child themes to take advantage of the accessible styles.

juanfra: Perfect.

carolinan: I did not understand 😄  Should this be updated to `get_stylesheet_uri` or not?

functions.php
carolinan: Thank you, fixed.

functions.php
carolinan: @justintadlock might have some input on the use of `get_parent_theme_file_uri`?

For Twenty Twenty-Four, there was feedback from users that they wanted the style.css to be enqueued because they often make changes to it (they shouldn't, but they do...).

justintadlock: @carolinan it looks good to me since you're intentionally loading the TT5 `style.css`, regardless of whether it or a child theme is active. 👍

carolinan: thank you

carolinan:
Eh no the contact pattern is not supposed to be on this branch o.o

PR #111, opened 2024-08-23
Add pattern: Contact - Location and directions link

### LINKED ISSUE #62: Block Patterns - Contact -  Location and directions link
**Description**

🛠️ The goal of this issue is to track the implementation of the `Contact - Location and directions link` pattern. 
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1815&t=IqDHO3nwfV13Hcvm-1)

Contact_ Location and directions link

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/62

**Screenshots**

https://github.com/user-attachments/assets/477bde7c-a3cf-444f-9af1-43db77b14a73

**Testing Instructions**

1. Create a page.
2. Add the pattern.
3. Confirm that it's looking like the designs (spacing/colors/image/etc).


huzaifaalmesbah:
looks good to me.

carolinan:
I updated the PR from trunk to be able to test the pattern in the current default templates.
There is some extra spacing on the right.
In the editor, this is only visible when both sidebars are closed.
I think this can be fixed by setting "Inner blocks use content width" to false on the outer group block.

kafleg (approved):
Design wise, it is looking good.

carolinan:
> Design wise, it is looking good.

Do you not see the extra spacing mentioned above? 🤔

juanfra:
Thanks for the reviews!

@carolinan, thanks for the heads-up. I was testing on my laptop, and now, with the external screen I can see what you highlighted. Can you please take a look at the changes? Thanks in advance.

carolinan:
By centering, the spacing on the sides is even, but it is also larger than expected.

juanfra:
@carolinan I can't get to see that. Can you please share a screenshot or screencast? 

From what I see in Figma, the pattern should have a layout where the max-width is "site wide". This means that if the site is viewed on large screens, there will be some space on both sides effectively.

carolinan:
That is the issue, the design in Figma clearly has a 50px padding around it, I am not going to be able to second guess the design and how it is intended to be implemented.

carolinan:
I do not see the pattern placed in any page design where it is not full width.

juanfra:
I believe the `50px` is the standard padding of these types of patterns on a full-width scenario, where the width goes to `1340px`. 

That's how I [understand it](https://github.com/WordPress/twentytwentyfive/pull/129#issuecomment-2312291056), and how it goes in line with how we implemented footers and headers.

@beafialho can you please look at this PR when you have time?

beafialho:
Thank you @juanfra, this pattern is looking good, I see no issues with it.

Issue #3, opened 2024-07-01
Section style slugs


richtabor:
With [section styles](https://make.wordpress.org/core/2024/06/24/section-styles/) landing in WordPress 6.6, let's talk about standardizing the `slug` values within Twenty Twenty Five. 

### Proposal 

I've experimented with [the Assembler theme](https://github.com/Automattic/themes/pull/7891), settling on `is-style-section-1`, `is-style-section-2`, slugs—as not every color variation has a "light" or "muted" or "dark" style (as some theme style variations are brighter, darker, or more minimal than others).

I've found that using a more generic slug will allow for section styles to be more meaning across multiple theme style variations—and also other themes—much like how spacing presets work. 

Twenty Twenty Five is a great opportunity to showcase color and typography, especially with section styles across many color variations; we need to account for the variety of color-only theme style variations that may be included in the theme. 

We could potentially leverage the pattern directory to make Twenty Twenty Five's patterns available for all WordPress installs leveraging the block editor, which would mean a standard for section styles could be interoperable across any site—no more colors that are fixed in a pattern.  

### Visual

Section styles, as applied to each pricing table here: 

![CleanShot 2024-07-01 at 14 17 04](https://github.com/WordPress/twentytwentyfive/assets/1813435/94f6cdd2-8100-4f4b-99fb-a34aa5c332cb)

richtabor:
Here's a case where the default muted, light, and dark section styles would map great—but if a color variation is much more expressive, then those slug values are much more confusing: 

https://github.com/WordPress/twentytwentyfive/assets/1813435/6bb493b8-5e7d-4988-8658-88e0606f1145

bgardner:
Noting this is also being discussed at: https://github.com/WordPress/gutenberg/issues/63036

bgardner:
I love the movement toward an interoperable (instead of opinionated) approach and want to adopt whatever we decide here/TT5 in my work with Powder Studio. I can't think of anything better (or more suitable) at the moment than the generic `section-1`, `section-2`, and `section-3` approach for slugs. This naming convention gets my vote.

justintadlock:
Agreed that setting a standard convention here with Twenty Twenty-Five would help the theming community in terms of interoperability and documentation. I'm in favor of `section-1`, `section-2`, etc. for these sort of generic sections (you can always change the label/title). I've been testing with these slugs over the past few weeks, and it's worked well across a variety of designs I have.

[...]

PR #118, opened 2024-08-24
Add patterns: Testimonials

### LINKED ISSUE #57: Block Patterns - Testimonials - Review with large image on the right
**Description**

🛠️ The goal of this issue is to track the implementation of the `Testimonial - Review with large image on the right` pattern. 
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1810&t=IqDHO3nwfV13Hcvm-1)

Testimonials_ Review with large image on the right

### LINKED ISSUE #58: Block Patterns - Testimonials - Two columns with avatar
**Description**

🛠️ The goal of this issue is to track the implementation of the `Testimonial - Two columns with avatar` pattern. 
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1811&t=IqDHO3nwfV13Hcvm-1)

Testimonials_ 2 columns with avatar

### LINKED ISSUE #59: Block Patterns - Testimonials - Three column layout with 6 testimonials
**Description**

🛠️ The goal of this issue is to track the implementation of the `Testimonial - Three column layout with 6 testimonials` pattern. 
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1812&t=IqDHO3nwfV13Hcvm-1)

Testimonials_ 3 column layout with 6 testimonials

g-elwell:
**Description**

Fixes #57, #58, #59

I've tackled all three of these in a single PR as working through them I realised they depend on one another.

- I've set the "default" quote block style with a border as seen in the 3 column layout. The "plain" styling removes this border/padding.
- The base block sizing/spacing is as per the 2 and 3 column layouts, the larger single testimonial has increased sizing/spacing applied to the block itself

This PR also registers the `Capsule` block style variation for `core/heading` blocks which is used in a few sections. It seems that a heading would be semantically correct in the locations where this is being used.

Couple of notes:

- ~~I'm not happy with the vertical alignment of the heading/quote. Unfortunately, `space-between` isn't an option on stack (only on row) which is a shame as this is all it would need to be perfect. Unless there's another way to do this I'm not aware of?~~ Edit: found this in the block toolbar!

- We're limited with how we can style the location part of the citation, so I'm using a line break and `subscript` to get this working. Not that intuitive to set up, but I think this is okay for a pre-made pattern?

- The Figma uses a different font just for the location part of the citation (instrument sans), I'm assuming this should be using manrope?

**Screenshots**

image

**Testing Instructions**

1. Create a page
1. Add the patterns (found under 'testimonials')
1. Confirm that they looks like the designs (spacing/colors/etc.)


g-elwell:
Apologies if anyone started reviewing this, as I started working on other testimonial patterns I realised they would be better within a single PR so have pushed several updates here and changed the title/description to reflect this.

styles/capsule.json
```diff
@@ -0,0 +1,23 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Capsule",
+	"slug": "capsule",
```
carolinan: We need to come up with a consistent system for the slugs and for organizing the style files in subfolders.
https://github.com/WordPress/twentytwentyfive/issues/3

I have already added a very similar style
https://github.com/WordPress/twentytwentyfive/blob/trunk/styles/blocks/post-terms-1.json

I wonder if we can use the same style for both blocks? I am not sure if the inconstancies in the border color is intentional, like it is meant to be different for the visual hierarchy? @beafialho

carolinan: Ah, I just remembered that the terms one need to style the **links** only, since there are multiple individual items. So while it will have very similar styling it needs to be separate style variations, unless we want to add **custom CSS** that can be reused for both.

g-elwell: Thanks for looking at this @carolinan - and for the resources on fullsiteediting.com as until I started this I wasn't even aware we could add style variations as json until I read it on there! 🙂 

It's a shame we can't reuse the same style without custom CSS. I assume we'd prefer to stick with style variations to push this as the recommended convention to use?

I think the subfolder organisation makes sense, but the slug naming could be tricky to apply to block styles. For example, if we opt for `heading-1` here, we're ignoring the fact that this style _could_ be reused, on paragraphs for example, by extending the `blockTypes` array.

I understand the portability aspect relating to section styles, but perhaps block style variations _should_ be named according to the type of styling they apply, as they're so specific. Clearly yourself and others have already been discussing/thinking about this at length, so I'm happy to make whatever changes you suggest here.

carolinan: I think this should not block the rest of the PR. Since these styles, for now, are not used in many templates or patterns, changing the slug later is not a blocker.

theme.json
```diff
@@ -1051,6 +1051,58 @@
 					"fontSize": "var:preset|font-size|x-large"
 				}
 			},
+			"core/quote": {
+				"border": {
```
juanfra: I believe these should not go into the quote core styles, as they're different. Please check the Figma designs for the quote [here](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=77-1417&t=iVc8k2Kl9C1qrITw-1).

beafialho:
Apologies for the delay, taking a look now!

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,43 @@
...
+<p class="has-x-large-font-size" style="line-height:1.1"><?php echo esc_html_x( '&ldquo;Superb product and customer service!&rdquo;', 'Sample testimonial', 'twentytwentyfive' ); ?></p>
+<!-- /wp:paragraph --></div>
+<!-- /wp:group --><cite><?php echo wp_kses_post( _x( 'Jo Mulligan <br /><sub>Atlanta, GA</sub>', 'Sample testimonial citation', 'twentytwentyfive' ) ); ?></cite></blockquote>
+<!-- /wp:quote --></div>
+<!-- /wp:group --></div>
+<!-- /wp:column -->
+
+<!-- wp:column {"style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"}}}} -->
+<div class="wp-block-column" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:group {"style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group" style="padding-top:0;padding-right:var(--wp--preset--spacing--50);padding-bottom:0;padding-left:var(--wp--preset--spacing--50)"><!-- wp:image {"aspectRatio":"1","scale":"cover","sizeSlug":"large","linkDestination":"none","align":"center"} -->
```
beafialho: Aspect ratio should be "Original", not Square.

beafialho: Here's what I'm seeing:

https://github.com/user-attachments/assets/a6eb8829-07fb-4e08-b38d-219fcbc0cbe8

beafialho: The image is set to "align center", it should be set to full to align with the wide width of the theme.

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,43 @@
...
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
+<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignwide" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)"><!-- wp:columns {"align":"wide"} -->
+<div class="wp-block-columns alignwide"><!-- wp:column -->
+<div class="wp-block-column"><!-- wp:group {"style":{"dimensions":{"minHeight":"100%"},"spacing":{"padding":{"top":"0","bottom":"0","left":"var:preset|spacing|50","right":"var:preset|spacing|50"},"blockGap":"var:preset|spacing|60"}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"stretch","flexWrap":"wrap","verticalAlignment":"space-between"}} -->
```
beafialho: The stack block containing the row and quote should not have left and right padding, they should align with the column edges to align with the rest of the content in the theme and patterns. Like I'm showing in the video:

https://github.com/user-attachments/assets/65024ee3-dd0a-4ffb-b768-f5c9dfad3207

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,43 @@
...
+ * Categories: testimonials
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
+<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
```
beafialho: Ideally, this outer group should be set to full width, so that if we add a background color it goes to the screen's edges.

https://github.com/user-attachments/assets/a244ea6b-e798-489c-9993-a201f0548d8f

patterns/testimonials-three-columns-six-testimonials.php
```diff
@@ -0,0 +1,83 @@
...
+ * Categories: testimonials
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
+<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"},"blockGap":"var:preset|spacing|50"}},"layout":{"type":"constrained"}} -->
```
beafialho: This group should be set to full, so that if we add a background color it looks good.
Captura de ecrã 2024-08-30, às 16 57 34

patterns/testimonials-two-columns-avatar.php
```diff
@@ -0,0 +1,61 @@
...
+ * Categories: testimonials
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
+<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
```
beafialho: This group should be set to full, so that if we add a background color it looks good.

Captura de ecrã 2024-08-30, às 16 58 05

beafialho (commented):
I left some comments directly in the code bits that need addressing. Thank you for your work on this.

patterns/testimonials-review-large-image-right.php
g-elwell: > Aspect ratio should be "Original", not Square.

Do you think we should crop the source image in that case, or are you happy with this deviating from the [figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1810&node-type=SYMBOL&t=6iAKPXq8Uv9G9ZbM-0)? It would look quite different with "original" aspect ratio applied:
image

> Here's what I'm seeing:
> 
>  image.mp4

This doesn't look right, I'm not seeing it locally (clean WP 6.6 with GB plugin enabled or disabled)
In your video it's visible that the image is overflowing the square bounding box, could something be interfering with it?

patterns/testimonials-review-large-image-right.php
g-elwell: Thanks, updated in https://github.com/WordPress/twentytwentyfive/pull/118/commits/6b65a0c983ace36d4f148cd61f8410f4b5ca5744

patterns/testimonials-review-large-image-right.php
g-elwell: Thanks, updated in https://github.com/WordPress/twentytwentyfive/pull/118/commits/6b65a0c983ace36d4f148cd61f8410f4b5ca5744

patterns/testimonials-two-columns-avatar.php
g-elwell: Thanks, updated in https://github.com/WordPress/twentytwentyfive/pull/118/commits/1c58d688ee227007220d325ba4b250926a49eb9f

patterns/testimonials-three-columns-six-testimonials.php
g-elwell: Thanks, updated in https://github.com/WordPress/twentytwentyfive/pull/118/commits/1467bbe13c4e8b970f30ca9aba2c23eee46e4009

g-elwell:
I've just pushed a commit https://github.com/WordPress/twentytwentyfive/pull/118/commits/79b45971ac774d07f67a17d2c2f77f1d4851989a which takes the standalone `core/quote` block styling into account as raised by @juanfra [here](https://github.com/WordPress/twentytwentyfive/pull/118#discussion_r1735915253). This meant adding more elements to this PR, but I feel having those base styles on the block is important so we know the pattern overrides are correct.

What I've ended up with looks like this:
image

Since the base block styles are fairly different to the quotes used across patterns, the patterns do have a fair amount of custom styling in them to get them to match the figma designs.

I've also included a custom style on the `core/columns` block for achieving the outline/border effect. Inside is a "plain" quote block. I feel this is probably more reusable for end-users, should they wish to have a similar column/box layout with different content.

If any of this seems incorrect please let me know!

theme.json
g-elwell: Thanks I didn't see that part of the figma. Have updated this and added a new comment below https://github.com/WordPress/twentytwentyfive/pull/118#issuecomment-2322611124

theme.json
```diff
@@ -1051,6 +1051,67 @@
...
+					"cite": {
+						"typography": {
+							"fontSize": "0.65em",
+							"fontStyle": "normal",
+							"fontWeight": "400"
+						},
+						"css": "& sub { font-size: 0.65em }"
+					}
+				},
+				"css": "&.has-text-align-right { border-left: 0; padding-left: 0; } &.has-text-align-center { border-inline: 0; padding-inline: 0; } &:not(.has-text-align-right):not(.has-text-align-center) { border-right: 0; padding-right: 0; }",
```
g-elwell, line 1226: This lets us achieve the different borders depending on the quote block alignment value. It takes into account left, right, center and no specified alignment

theme.json
```diff
@@ -1051,6 +1051,67 @@
...
+					"fontWeight": "600"
+				},
+				"elements": {
+					"cite": {
+						"typography": {
+							"fontSize": "0.65em",
+							"fontStyle": "normal",
+							"fontWeight": "400"
+						},
+						"css": "& sub { font-size: 0.65em }"
```
g-elwell: I would have preferred to use preset values here, but these seem to need to scale with the base font size of the block based on the designs.

patterns/testimonials-review-large-image-right.php
beafialho: Apologies, ignore my previous comment about the aspect ratio. Let's make sure it's **square**, but it's not looking zoomed in like my recording above and it looks like a bug.

carolinan:
Hi!
Please update the PR to include the latest changes from trunk.

I did not understand why the style variation for the columns is added, I don't think it is needed.
With Gutenberg active, and in WordPress 6.7, the columns will have radius added to the border controls, so users can already add this style without too much effort.

The license / credit and source information for the avatar image must be added to the readme.txt file. Since this avatar is an example that will be small, I think the width and height of the image can be reduced drastically and still look acceptable.

The image of the miner is already in the theme, the name of the image file is team_photo_3.webp.

carolinan:
The capsule style also needs to be replaced with the existing pill style.

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,39 @@
...
+ * Categories: testimonials
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
```
juanfra: Can you please set the margin to zero for the outer group so that when a style variation is added to the group (once trunk is merged to the PR) there's no gap between groups?

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,39 @@
...
+<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)"><!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|60"}}}} -->
+<div class="wp-block-columns alignwide"><!-- wp:column -->
+<div class="wp-block-column"><!-- wp:group {"style":{"dimensions":{"minHeight":"100%"},"spacing":{"blockGap":"var:preset|spacing|60"}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"left","flexWrap":"wrap","verticalAlignment":"space-between"}} -->
+<div class="wp-block-group" style="min-height:100%"><!-- wp:heading {"className":"is-style-capsule","style":{"layout":{"selfStretch":"fit","flexSize":null}},"fontSize":"x-small"} -->
```
juanfra: Can we use the [pill style](https://github.com/WordPress/twentytwentyfive/pull/210) variation we have already in the theme?

Once we use the pill style we should remove the "capsule" variation in favor of the "pill" variation.

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,39 @@
...
+<h2 class="wp-block-heading is-style-capsule has-x-small-font-size"><?php echo esc_html_x( 'What people are saying', 'Testimonial heading', 'twentytwentyfive' ); ?></h2>
+<!-- /wp:heading -->
+
+<!-- wp:quote {"className":"is-style-plain","style":{"spacing":{"blockGap":"var:preset|spacing|50"},"typography":{"fontStyle":"normal","fontWeight":"400"}},"fontSize":"large"} -->
+<blockquote class="wp-block-quote is-style-plain has-large-font-size" style="font-style:normal;font-weight:400"><!-- wp:group {"style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained","justifyContent":"left","contentSize":"400px"}} -->
+<div class="wp-block-group" style="margin-top:0;margin-bottom:0;padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.1"}},"fontSize":"x-large"} -->
```
juanfra: Is there any reason why you are wrapping the quote inside a group? We should be good without it.

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,39 @@
...
+<div class="wp-block-column"><!-- wp:group {"style":{"dimensions":{"minHeight":"100%"},"spacing":{"blockGap":"var:preset|spacing|60"}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"left","flexWrap":"wrap","verticalAlignment":"space-between"}} -->
+<div class="wp-block-group" style="min-height:100%"><!-- wp:heading {"className":"is-style-capsule","style":{"layout":{"selfStretch":"fit","flexSize":null}},"fontSize":"x-small"} -->
+<h2 class="wp-block-heading is-style-capsule has-x-small-font-size"><?php echo esc_html_x( 'What people are saying', 'Testimonial heading', 'twentytwentyfive' ); ?></h2>
+<!-- /wp:heading -->
+
+<!-- wp:quote {"className":"is-style-plain","style":{"spacing":{"blockGap":"var:preset|spacing|50"},"typography":{"fontStyle":"normal","fontWeight":"400"}},"fontSize":"large"} -->
+<blockquote class="wp-block-quote is-style-plain has-large-font-size" style="font-style:normal;font-weight:400"><!-- wp:group {"style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained","justifyContent":"left","contentSize":"400px"}} -->
```
juanfra: Is there any particular reason to use the content size? We should be good without it.

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,39 @@
...
+<p class="has-x-large-font-size" style="line-height:1.1"><?php echo esc_html_x( '&ldquo;Superb product and customer service!&rdquo;', 'Sample testimonial', 'twentytwentyfive' ); ?></p>
+<!-- /wp:paragraph --></div>
+<!-- /wp:group --><cite><?php echo wp_kses_post( _x( 'Jo Mulligan <br /><sub>Atlanta, GA</sub>', 'Sample testimonial citation', 'twentytwentyfive' ) ); ?></cite></blockquote>
+<!-- /wp:quote --></div>
+<!-- /wp:group --></div>
+<!-- /wp:column -->
+
+<!-- wp:column {"style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"}}}} -->
+<div class="wp-block-column" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:image {"aspectRatio":"1","scale":"cover","sizeSlug":"large","linkDestination":"none","align":"center"} -->
+<figure class="wp-block-image aligncenter size-large"><img src="<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/3906400735_8f0fa9231d_o.webp" alt="<?php echo esc_attr_x( 'Picture of a person', 'Alt text for testimonial image', 'twentytwentyfive' ); ?>" style="aspect-ratio:1;object-fit:cover"/></figure>
```
juanfra: We need to add the license of the images to the `readme.txt` file.

patterns/testimonials-review-large-image-right.php
```diff
@@ -0,0 +1,39 @@
...
+ * Keywords: testimonial
+ * Categories: testimonials
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
```
juanfra: Can we remove this empty line before the group?

patterns/testimonials-three-columns-six-testimonials.php
```diff
@@ -0,0 +1,83 @@
...
+ * Keywords: testimonial
+ * Categories: testimonials
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
```
juanfra: Can we remove this empty line before the group?

patterns/testimonials-three-columns-six-testimonials.php
```diff
@@ -0,0 +1,83 @@
...
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"},"blockGap":"var:preset|spacing|50"}},"layout":{"type":"constrained"}} -->
```
juanfra: Can you please set the margin to zero for the outer group so that when a style variation is added to the group (once trunk is merged to the PR) there's no gap between groups?

patterns/testimonials-three-columns-six-testimonials.php
```diff
@@ -0,0 +1,83 @@
...
+<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)"><!-- wp:heading {"align":"wide","fontSize":"xx-large"} -->
+<h2 class="wp-block-heading alignwide has-xx-large-font-size"><?php echo esc_html_x( 'What people are saying', 'Testimonial section heading', 'twentytwentyfive' ); ?></h2>
+<!-- /wp:heading -->
+
+<!-- wp:columns {"align":"wide","className":"is-style-columns-1","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|60","left":"var:preset|spacing|50"}}}} -->
+<div class="wp-block-columns alignwide is-style-columns-1"><!-- wp:column -->
```
juanfra: We should be able to set the styles in columns without defining a new preset for columns. In order to do that, you should be using the latest Gutenberg.

patterns/testimonials-three-columns-six-testimonials.php
```diff
@@ -0,0 +1,83 @@
...
+<div class="wp-block-column"><!-- wp:quote {"className":"is-style-plain","style":{"typography":{"fontStyle":"normal","fontWeight":"400"},"spacing":{"blockGap":"var:preset|spacing|40"}},"fontSize":"large"} -->
+<blockquote class="wp-block-quote is-style-plain has-large-font-size" style="font-style:normal;font-weight:400"><!-- wp:group {"style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained","justifyContent":"left","contentSize":"400px"}} -->
+<div class="wp-block-group" style="margin-top:0;margin-bottom:0;padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.1"}}} -->
```
juanfra, line 23: Is there any reason why we are using a group to wrap the content of the quote? We should be good without it. I also believe there's no need to have the content size to 400px. This applies to all of the quotes in this pattern.

patterns/testimonials-two-columns-avatar.php
```diff
@@ -0,0 +1,61 @@
...
+ * Keywords: testimonial
+ * Categories: testimonials
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
```
juanfra: Can you please remove this empty line before the group?

patterns/testimonials-two-columns-avatar.php
```diff
@@ -0,0 +1,61 @@
...
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
```
juanfra: Can you please set the margin to zero for the outer group so that when a style variation is added to the group (once trunk is merged to the PR) there's no gap between groups?

patterns/testimonials-two-columns-avatar.php
```diff
@@ -0,0 +1,61 @@
...
+<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)"><!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|60","left":"var:preset|spacing|50"}}}} -->
+<div class="wp-block-columns alignwide"><!-- wp:column -->
+<div class="wp-block-column"><!-- wp:columns {"style":{"spacing":{"blockGap":{"top":"var:preset|spacing|40","left":"var:preset|spacing|40"}}}} -->
+<div class="wp-block-columns"><!-- wp:column {"width":"64px"} -->
+<div class="wp-block-column" style="flex-basis:64px"><!-- wp:image {"width":"64px","aspectRatio":"1","scale":"cover","sizeSlug":"large","linkDestination":"none","className":"is-style-rounded"} -->
+<figure class="wp-block-image size-large is-resized is-style-rounded"><img src="<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/image-from-rawpixel-id-5918459-original.webp" alt="<?php echo esc_attr_x( 'Picture of a person', 'Alt text for testimonial image', 'twentytwentyfive' ); ?>" style="aspect-ratio:1;object-fit:cover;width:64px"/></figure>
```
juanfra: Can you please add the license for the image in the `readme.txt` file?

patterns/testimonials-two-columns-avatar.php
```diff
@@ -0,0 +1,61 @@
...
+<!-- /wp:image --></div>
+<!-- /wp:column -->
+
+<!-- wp:column -->
+<div class="wp-block-column"><!-- wp:quote {"className":"is-style-plain","style":{"typography":{"fontStyle":"normal","fontWeight":"400"},"spacing":{"blockGap":"var:preset|spacing|40"}},"fontSize":"large"} -->
```
juanfra: Like the other pattern, the styles for the border of the columns shouldn't be coming from the variation but they could go directly in the column.

patterns/testimonials-two-columns-avatar.php
```diff
@@ -0,0 +1,61 @@
...
+<blockquote class="wp-block-quote is-style-plain has-large-font-size" style="font-style:normal;font-weight:400"><!-- wp:group {"style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained","justifyContent":"left","contentSize":"400px"}} -->
+<div class="wp-block-group" style="margin-top:0;margin-bottom:0;padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:paragraph {"style":{"typography":{"lineHeight":"1.1"}}} -->
```
juanfra: Is there any reason why we are using a group to wrap the content of the quote? We should be good without it. I also believe there's no need to have the content size to 400px. This applies to all of the quotes in this pattern.

styles/blocks/capsule.json
```diff
@@ -0,0 +1,23 @@
+{
```
juanfra: Thanks for the work here, but we already have a "pill" style defined for this. Let's go with [that one](https://github.com/WordPress/twentytwentyfive/pull/210) instead of capsule.

styles/blocks/columns-1.json
```diff
@@ -0,0 +1,28 @@
+{
```
juanfra: I believe once we apply the styles in the columns directly we shouldn't need this one.

theme.json
```diff
@@ -1051,6 +1051,67 @@
 					"fontSize": "var:preset|font-size|x-large"
 				}
 			},
+			"core/quote": {
```
juanfra: Can we please rebase, and check how these styles live together with the work already done in the `theme.json`? There may be some that can be redundant after that.

juanfra (commented):
Thank you so much for your work on this @g-elwell, and your patience.

Can you please rebase and check the comments we left for this PR? Thanks again! 🙏

patterns/testimonials-review-large-image-right.php
g-elwell: Resolved via https://github.com/WordPress/twentytwentyfive/pull/118/commits/7071583bedf0e7cd56e65e9e58229897b225a33c

patterns/testimonials-review-large-image-right.php
g-elwell: Resolved via https://github.com/WordPress/twentytwentyfive/pull/118/commits/910704280fa23a8ba1c2ccff154dde8bd70feb7a

patterns/testimonials-review-large-image-right.php
g-elwell: Sure, resolved in https://github.com/WordPress/twentytwentyfive/pull/118/commits/953bec6e6a4920d3608bc745e5c088c51dd96028

patterns/testimonials-review-large-image-right.php
g-elwell: This (and the use of a group) is to ensure the text wraps correctly in line with the design - unless there's a better way I'm not aware of? 

https://github.com/user-attachments/assets/ff41032a-198e-4f7d-838b-868be6583af3

patterns/testimonials-review-large-image-right.php
g-elwell: See https://github.com/WordPress/twentytwentyfive/pull/118/files#r1755578274

patterns/testimonials-three-columns-six-testimonials.php
g-elwell: Resolved in https://github.com/WordPress/twentytwentyfive/pull/118/commits/1003c19e9762a28a7a6e263016b851dd796f9a8a

patterns/testimonials-three-columns-six-testimonials.php
g-elwell: Resolved in https://github.com/WordPress/twentytwentyfive/pull/118/commits/ed6cfa465af74099c826627c7dc08ce3cf1945c2

patterns/testimonials-review-large-image-right.php
g-elwell: I've replaced this image with one already added to theme (which has licence added in readme.txt) https://github.com/WordPress/twentytwentyfive/pull/118/commits/264b44a5cabee7d86828120cb9c359b12edc7616

patterns/testimonials-three-columns-six-testimonials.php
g-elwell: Nice! Have updated pattern styles and removed the unneeded variation in https://github.com/WordPress/twentytwentyfive/pull/118/commits/a580f0c6ccd69d21d15519c2a191a1a2db413530

patterns/testimonials-three-columns-six-testimonials.php
g-elwell: This is just to prevent overly long lines of text when the columns stack:

image

The width is set to match how the other quote patterns wrap their text in the design:

image

patterns/testimonials-two-columns-avatar.php
g-elwell: Resolved via https://github.com/WordPress/twentytwentyfive/pull/118/commits/7a24051a4e022049c57762ac8efbce1a84b27913

patterns/testimonials-two-columns-avatar.php
g-elwell: Resolved in https://github.com/WordPress/twentytwentyfive/pull/118/commits/78363e7ac7f8581fcc3d07596419be5b16d12022

styles/blocks/columns-1.json
g-elwell: Removed in https://github.com/WordPress/twentytwentyfive/pull/118/commits/a580f0c6ccd69d21d15519c2a191a1a2db413530

styles/blocks/capsule.json
g-elwell: Switched to the pill style in https://github.com/WordPress/twentytwentyfive/pull/118/commits/953bec6e6a4920d3608bc745e5c088c51dd96028

patterns/testimonials-two-columns-avatar.php
g-elwell: Similar to the other areas where I've taken this approach, the reason is to prevent long text lines at mid screen widths - this is more noticeable when longer text is used in a quote, see:

image

Happy to remove if you think it's not needed, but that was the reason I went for it 🙂

patterns/testimonials-two-columns-avatar.php
g-elwell: Here, the style is on the quote block. The default quote is styled to have a border, so I added a 'plain' variation which removes the border and is used in all of the patterns.

patterns/testimonials-two-columns-avatar.php
g-elwell: Added in https://github.com/WordPress/twentytwentyfive/pull/118/commits/40cc4b09c36aee39a342b97dc907ded1580271a5

g-elwell:
> Hi! Please update the PR to include the latest changes from trunk.

Done, thank you!

> I did not understand why the style variation for the columns is added, I don't think it is needed. With Gutenberg active, and in WordPress 6.7, the columns will have radius added to the border controls, so users can already add this style without too much effort.

Removed in https://github.com/WordPress/twentytwentyfive/pull/118/commits/a580f0c6ccd69d21d15519c2a191a1a2db413530

> The license / credit and source information for the avatar image must be added to the readme.txt file. Since this avatar is an example that will be small, I think the width and height of the image can be reduced drastically and still look acceptable.

Updated in https://github.com/WordPress/twentytwentyfive/pull/118/commits/40cc4b09c36aee39a342b97dc907ded1580271a5

> The image of the miner is already in the theme, the name of the image file is team_photo_3.webp.

I've replaced the image with the one already merged into the theme (https://github.com/WordPress/twentytwentyfive/pull/118/commits/264b44a5cabee7d86828120cb9c359b12edc7616), however I suggest we crop the image to remove the black border (I had done this previously on my version). Currently it looks like this:
image

> The capsule style also needs to be replaced with the existing pill style.

Updated in https://github.com/WordPress/twentytwentyfive/pull/118/commits/953bec6e6a4920d3608bc745e5c088c51dd96028

g-elwell:
I've worked through all the feedback given so far. I also noticed that some upstream font sizes affected these patterns, so needed to bump the quote font sizes up to `x-large`. In the case of the `Review with large image on right` pattern, I found `x-large` to be to small and `xx-large` to be way too large, so used a one-off font size there. Current screenshot:

image

patterns/testimonials-review-large-image-right.php
juanfra: Thanks!

beafialho:
Unfortunately I'm not able to test this because I'm seeing a lot of conflicts with `theme.json` on my GH desktop from the latest commits. From the screenshots, this looks fine but I'd like to be able to test it. @g-elwell could you please try to fix these conflicts?

patterns/testimonials-two-columns-avatar.php
juanfra: Thanks for the clarification, @g-elwell. I see this is an edge case likely to happen on tablets in portrait mode, and your approach can solve it.

I'll defer to @beafialho so that she can give her input on what's best for the design.

patterns/testimonials-two-columns-avatar.php
juanfra: I see. Sounds good, thanks for the clarification.

patterns/testimonials-two-columns-avatar.php
juanfra: The decision will apply to these other comments:

- https://github.com/WordPress/twentytwentyfive/pull/118#discussion_r1755687401
- https://github.com/WordPress/twentytwentyfive/pull/118#discussion_r1755578274

Thanks ✨

patterns/testimonials-three-columns-six-testimonials.php
juanfra: Thanks, I replied [here](https://github.com/WordPress/twentytwentyfive/pull/118/files#r1756563819) so that we can centralize the discussion.

patterns/testimonials-review-large-image-right.php
juanfra: Thanks, I replied [here](https://github.com/WordPress/twentytwentyfive/pull/118/files#r1756563819) so that we can centralize the discussion.

patterns/testimonials-review-large-image-right.php
juanfra: Thanks, I replied [here](https://github.com/WordPress/twentytwentyfive/pull/118/files#r1756563819) so that we can centralize the discussion.

patterns/testimonials-three-columns-six-testimonials.php
```diff
@@ -0,0 +1,82 @@
...
+<h2 class="wp-block-heading alignwide has-xx-large-font-size"><?php echo esc_html_x( 'What people are saying', 'Testimonial section heading', 'twentytwentyfive' ); ?></h2>
+<!-- /wp:heading -->
+
+<!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|60","left":"var:preset|spacing|50"}}}} -->
+<div class="wp-block-columns alignwide"><!-- wp:column {"style":{"border":{"width":"1px","color":"#11111133","radius":"10px"},"spacing":{"padding":{"right":"var:preset|spacing|40","left":"var:preset|spacing|40","top":"var:preset|spacing|40","bottom":"var:preset|spacing|40"}}}} -->
+<div class="wp-block-column has-border-color" style="border-color:#11111133;border-width:1px;border-radius:10px;padding-top:var(--wp--preset--spacing--40);padding-right:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--40);padding-left:var(--wp--preset--spacing--40)"><!-- wp:quote {"className":"is-style-plain","style":{"typography":{"fontStyle":"normal","fontWeight":"400"},"spacing":{"blockGap":"var:preset|spacing|40"}},"fontSize":"x-large"} -->
```
juanfra: Let's use the opacity preset for the border color, here and in the other appearances.

g-elwell:
> Unfortunately I'm not able to test this because I'm seeing a lot of conflicts with `theme.json` on my GH desktop from the latest commits. From the screenshots, this looks fine but I'd like to be able to test it. @g-elwell could you please try to fix these conflicts?

Hi @beafialho - the conflicts should be resolved now if you'd like to give this another try? Thanks!

juanfra (commented):
Thanks for the hard work on this one @g-elwell - There's a lot of progress and we're getting closer.

> I've replaced the image with the one already merged into the theme (https://github.com/WordPress/twentytwentyfive/commit/264b44a5cabee7d86828120cb9c359b12edc7616), however I suggest we crop the image to remove the black border (I had done this previously on my version). Currently it looks like this:

Let's follow your suggestion, as we need to have it without the border. Let's use the one I'm adding to this comment.

[image-from-rawpixel-id-5918459.webp.zip](https://github.com/user-attachments/files/16977545/image-from-rawpixel-id-5918459.webp.zip)

> I've worked through all the feedback given so far. I also noticed that some upstream font sizes affected these patterns, so needed to bump the quote font sizes up to x-large.

Since yesterday, we also have [custom font sizes](https://github.com/WordPress/twentytwentyfive/pull/269) for headings that you could use in patterns.

I see that there are some merge conflicts with the `theme.json`, can you please take a look and that? Ideally we should respect the format that's coming from trunk.

beafialho:
> Hi @beafialho - the conflicts should be resolved now if you'd like to give this another try? Thanks!

I'm still seeing these conflicts and I can't test it.

Captura de ecrã 2024-09-12, às 11 28 39

g-elwell:
> > Hi @beafialho - the conflicts should be resolved now if you'd like to give this another try? Thanks!
> 
> I'm still seeing these conflicts and I can't test it.
> 
> Captura de ecrã 2024-09-12, às 11 28 39

That's odd, it says it's resolved here and in the CLI. I'll install GH desktop and see what happens. Please bare with me.

patterns/testimonials-three-columns-six-testimonials.php
g-elwell: Done this in https://github.com/WordPress/twentytwentyfive/pull/118/commits/59e8dbdaaa2b896b36e3b9c7382565faf0df0b8a thanks

g-elwell:
Should be good now @beafialho 🤞 

> Thanks for the hard work on this one @g-elwell - There's a lot of progress and we're getting closer.
> 
> > I've replaced the image with the one already merged into the theme ([264b44a](https://github.com/WordPress/twentytwentyfive/commit/264b44a5cabee7d86828120cb9c359b12edc7616)), however I suggest we crop the image to remove the black border (I had done this previously on my version). Currently it looks like this:
> 
> Let's follow your suggestion, as we need to have it without the border. Let's use the one I'm adding to this comment.
> 
> [image-from-rawpixel-id-5918459.webp.zip](https://github.com/user-attachments/files/16977545/image-from-rawpixel-id-5918459.webp.zip)
> 

Done thanks https://github.com/WordPress/twentytwentyfive/pull/118/commits/19d738d218793114528edb5a548bf2e9d1282d7e

> > I've worked through all the feedback given so far. I also noticed that some upstream font sizes affected these patterns, so needed to bump the quote font sizes up to x-large.
> 
> Since yesterday, we also have [custom font sizes](https://github.com/WordPress/twentytwentyfive/pull/269) for headings that you could use in patterns.

Is there somewhere you'd prefer me to use these over the existing sizes? `x-large` seems to do the trick on most of these patterns. In the `Review with large image on right` pattern, it looks like it falls in between `heading-1` and `huge` so may need to remain a one-off?

beafialho:
I was able to test already, perhaps it was something particular with my environment. Thank you for the work done in these patterns so far!

I'm seeing a few issues:

**Review with large image on right**

- The image is overflowing the square edges. This is happening because the image is set to `align: center` and it should be `align: none`. The image should also not have the black border, it should look like the crop on Figma.

Captura de ecrã 2024-09-12, às 14 27 30

- The “Superb product and customer service!” should be set to `XXL` instead of `XL`.
- The "Jo Mulligan / Atlanta, GA" text is looking very tiny. This seems to have a specific quote in there, I don't know why but regardless, we should set it to `XL` instead of `L` so it is coherent with the other patterns.

https://github.com/user-attachments/assets/f817faa8-d3f5-4d8b-b8bb-d9001399e673



**Two columns with avatar**

- This pattern is also not looking good on tablet. If we untoggle the "Inner blocks use content width" for the group inside the Quote block, it looks much better in all screen sizes.

Captura de ecrã 2024-09-12, às 14 35 04

**Three column layout with 6 testimonials**

This one is looking good ✅

g-elwell:
> I was able to test already, perhaps it was something particular with my environment. Thank you for the work done in these patterns so far!
> 
> I'm seeing a few issues:
> 
> **Review with large image on right**
> 
> * The image is overflowing the square edges. This is happening because the image is set to `align: center` and it should be `align: none`. The image should also not have the black border, it should look like the crop on Figma.

I've removed the alignment and cropped the image

> 
> * The “Superb product and customer service!” should be set to `XXL` instead of `XL`.
> * The "Jo Mulligan / Atlanta, GA" text is looking very tiny. This seems to have a specific quote in there, I don't know why but regardless, we should set it to `XL` instead of `L` so it is coherent with the other patterns.

I've increased font sizes as suggested

> **Two columns with avatar**
> 
> * This pattern is also not looking good on tablet. If we untoggle the "Inner blocks use content width" for the group inside the Quote block, it looks much better in all screen sizes.

I've removed the group to remove the content width on this pattern

> **Three column layout with 6 testimonials**
> 
> This one is looking good ✅

Thanks!

Current screenshot:
image

theme.json
```diff
@@ -1192,26 +1192,53 @@
...
-					"fontStyle": "normal",
+					"fontSize": "var:preset|font-size|medium",
+					"lineHeight": "1.1",
 					"fontWeight": "600"
 				},
 				"elements": {
 					"cite": {
 						"typography": {
-							"fontSize": "var:preset|font-size|small",
+							"fontSize": "0.65em",
```
carolinan, line 1219: Why was this size changed again, why is  it using `em` and not `rem`?
It is an actual question, not a complaint or request 😄

carolinan: One would think the "task force" that has updated a large number of quote blocks in the old bundled themes the last 6 months would know why 😆 
It's so that the cite changes size relatively to whatever size the user sets on the quote.

carolinan:
I will merge this and then create a follow up for the duplicate images.

PR #120, opened 2024-08-25
Add Pattern: About Details Descriptions With Bullet

### LINKED ISSUE #66: Block Patterns - About - Detailed description with bullets
**Description**

🛠️ The goal of this issue is to track the implementation of the `About - Detailed description with bullets` pattern. 
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1768&t=IqDHO3nwfV13Hcvm-1)

About_ Detailed description with bullets

shail-mehta:
Fixes: https://github.com/WordPress/twentytwentyfive/issues/66

## Screenshot: 

about-details-descriptions-with-bullets

## Testing Instructions

- Create a page.
- Add the pattern.
- Confirm that it's looking like the designs (spacing/colors/text etc).

### DIFF
--- a/patterns/about-details-description-with-bullets.php
+++ b/patterns/about-details-description-with-bullets.php
@@ -12,8 +12,8 @@
 
 ?>
 
-<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70","left":"var:preset|spacing|70","right":"var:preset|spacing|70"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained","contentSize":""}} -->
-<div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--70);padding-right:var(--wp--preset--spacing--70);padding-bottom:var(--wp--preset--spacing--70);padding-left:var(--wp--preset--spacing--70)">
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70","left":"var:preset|spacing|80","right":"var:preset|spacing|80"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained","contentSize":""}} -->
+<div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--70);padding-right:var(--wp--preset--spacing--80);padding-bottom:var(--wp--preset--spacing--70);padding-left:var(--wp--preset--spacing--80)">
 	<!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|70","left":"var:preset|spacing|70"}}}} -->
 	<div class="wp-block-columns alignwide">
 		<!-- wp:column {"verticalAlignment":"top","width":"50%"} -->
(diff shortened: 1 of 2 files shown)


patterns/about-details-description-with-bullets.php
```diff
@@ -0,0 +1,75 @@
+<?php
+/**
+ * Title: About Details Description
+ * Slug: twentytwentyfive/about-details-description
+ * Title: About - Detailed description with bullets
+ * Slug: twentytwentyfive/about-details-description-with-bullets
```
carolinan, line 4: There should only be one slug and one title.

shail-mehta: Thanks for the Feedback.

There was a conflict with a previous implementation in the branch, but it has been resolved. I have removed the duplicate code to address the issue.

patterns/about-details-description-with-bullets.php
```diff
@@ -0,0 +1,63 @@
+<?php
+/**
+ * Title: About Details Description With Bullets
+ * Slug: twentytwentyfive/about-details-description-with-bullets
+ * Categories: text, about, columns
+ * Description: About Details Description With Bullet.
```
carolinan: Having the title and the description be the same text is not helpful.
From these texts, I can't understand what the pattern does.

patterns/about-details-description-with-bullets.php
```diff
@@ -0,0 +1,63 @@
...
+			<h3 class="wp-block-heading has-x-large-font-size">
+				<?php esc_html_e( 'Let’s hear about history and ancestry', 'twentytwentyfive' ); ?>
+			</h3>
+			<!-- /wp:heading -->
+		</div>
+		<!-- /wp:column -->
+
+		<!-- wp:column {"verticalAlignment":"center","width":"50%"} -->
+		<div class="wp-block-column is-vertically-aligned-center" style="flex-basis:50%">
+			<!-- wp:paragraph {"style":{"typography":{"textTransform":"uppercase"}},"textColor":"primary","fontSize":"small"} -->
```
carolinan: Since this describes the list below it, this should also be a heading. If the first is an H3, this needs to be an H4 😄

theme.json
```diff
@@ -1105,6 +1105,9 @@
 						}
 					}
 				}
+			},
+			"core/list": {
+				"css": "& {list-style-type:disc;} & li{margin-bottom: 0.5rem;}"
 			}
```
carolinan, line 1111: The list style type needs to be removed because it overrides the user's block setting.
Users need to be able to set the list to _ordered_ which is meant to use the numbers as the list style type.

shail-mehta:
Thanks for the feedback. 

I have addressed all the suggested changes.

carolinan:
Thank you.
The name still confuses me a bit because I am expecting the details block 🤷‍♀️

shail-mehta:
Could we use a title pattern such as `About Details Block with Bullets` or `Details Block with Bullets` ?

carolinan:
I don't understand... it does **not** use the details block.. its two headings and a list... ? That is why the name is confusing. I would expect "details" to be used if the pattern has a details block.

shail-mehta:
Thank you for your clarification. The title was based on the pattern name. 

However, I agree that Pattern Title **`Two Headings and a List`** more accurately reflects the pattern’s content. I will update the pattern title, slug, and file name to align with this description.

patterns/about-details-description-with-bullets.php
```diff
@@ -0,0 +1,63 @@
...
+?>
+
+<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70","left":"var:preset|spacing|80","right":"var:preset|spacing|80"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained","contentSize":""}} -->
+<div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--70);padding-right:var(--wp--preset--spacing--80);padding-bottom:var(--wp--preset--spacing--70);padding-left:var(--wp--preset--spacing--80)">
+	<!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|70","left":"var:preset|spacing|70"}}}} -->
+	<div class="wp-block-columns alignwide">
+		<!-- wp:column {"verticalAlignment":"top","width":"50%"} -->
+		<div class="wp-block-column is-vertically-aligned-top" style="flex-basis:50%">
+			<!-- wp:heading {"level":3,"fontSize":"x-large"} -->
+			<h3 class="wp-block-heading has-x-large-font-size">
```
juanfra, line 22: This should be an `H3` of `38px`, so we should remove the `x-large`.

shail-mehta:
Thank you both for the feedback.

I have created a new PR to address both changes.

PR #123, opened 2024-08-26
Add patterns for the Photo blog

### LINKED ISSUE #41: Add patterns for Photo blog
Figma link:
https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3155-1263&t=SbPR4qKXgP7EK643-4

Patterns:
All patterns should have "Inserter" set to "no" in the pattern file header.
The **template** patterns use the default header and footer template parts.

**Suggested** names and slugs:

- [x] Photo Home: template-home-photo.php  * Template Types: front-page, home
- [x] Photo Archive: template-archive-photo.php  * Template Types: archive
- [x] Photo Search: template-search-photo.php  * Template Types: search
- [x] Photo Single: template-single-photo.php  * Template Types: posts, single
- [x] Photo Page: template-page-photo.php

carolinan:
**Description**

Adds <del>templates and</del> patterns for the photo blog.
Closes https://github.com/WordPress/twentytwentyfive/issues/41

[Figma link](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3155-1263&t=4ZvvKDNEjcR8g2uv-4)

**Testing Instructions**
 On your WordPress test installation, create enough posts for the query pagination and next and previous posts to work:
 Create posts with and without featured images, with and without tags, and with no or several categories.
 
 Create two pages, one with and one without featured image.
 
Apply the PR and activate the theme.
Activate the Gutenberg plugin.

Go to Appearance > Editor and select the home template.
Open the Design panel and switch to "photo blog home template" (this needs a better shorter name).
Confirm if the design matches the design in Figma.
Save and view the the front, and confirm if it looks correct on the front.

Now repeat the test for the other templates: Archive, search, single post, page.

 
**Screenshots**

(out of date screenshots deleted)


carolinan:
The number of group blocks it took to create the header of the single post template seems unreasonable, it might be better to try to use a columns block.

patterns/template-home-photo.php
```diff
@@ -0,0 +1,32 @@
...
+?>
+
+<!-- wp:template-part {"slug":"header","area":"header","tagName":"header"} /-->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
+<main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
+	<!-- wp:heading {"textAlign":"center","level":1,"fontSize":"x-small"} -->
+	<h1 class="wp-block-heading has-text-align-center has-x-small-font-size">Stories</h1>
+	<!-- /wp:heading -->
+
```
carolinan: I did not add the style for this smaller heading, as it is already created in one of the other open pull requests:
https://github.com/WordPress/twentytwentyfive/pull/118

carolinan:
On smaller screens, the spacing around the image in the query loop is too large, but I was not able to make it work on the desktop sized screen without it:

mobile-home-photo

patterns/template-single-photo.php
```diff
@@ -0,0 +1,90 @@
...
+				</div>
+				<!-- /wp:group -->
+			</div>
+			<!-- /wp:group -->
+		</div>
+		<!-- /wp:group -->
+
+		<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"right":"var:preset|spacing|50","left":"var:preset|spacing|50"},"margin":{"top":"var:preset|spacing|50","bottom":"0"}}},"layout":{"type":"default"}} -->
+		<div class="wp-block-group alignfull" style="margin-top:var(--wp--preset--spacing--50);margin-bottom:0;padding-right:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)">
+			<!-- wp:group {"tagName":"nav","align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|40","bottom":"var:preset|spacing|40"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
```
carolinan: I tried setting the small font size directly on the group, but it did not work, the post navigation link did not inherit the size from the group. So the font size is added once for each block.

g-elwell:
> Home, front view home-photo

Unfortunately, this doesn't achieve the masonry effect from the Figma designs, as new rows don't stack up against previous items. Masonry is going to be tricky to pull off without JS, but the following as a style variation might be worth a try:
```css
{
	"version": 3,
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"title": "Masonry",
	"slug": "post-template-1",
	"blockTypes": ["core/post-template"],
	"styles": {
		"css": "display: block !important; column-count: 3;"
	}
}
```
Note, this will cause the post order to flow down columns rather than across rows as you'd expect, and it's also not responsive.

Bearing in mind users could render any combination of aspect ratio featured images here, if we want to provide a more predictable layout, we might want to consider enforcing 1:1 aspect ratio on the images, similar to the archive page.

beafialho:
Thank you for working on this. A few comments:

- Remove the dot in "Tell your story" (I know it had it on Figma, but I just removed it there as well)
- Center the "No posts found" text
- The query loop and content of the patterns should have wide width
- And an adjustment because unfortunately we can't have the masonry effect and it's not looking great, setting the featured images with square 1:1 aspect ratio like the [Archive and Search](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3137-2610&t=bCFceo6KG1JJBgYQ-4)

carolinan:
> * Remove the dot in "Tell your story" (I know it had it on Figma, but I just removed it there as well)

The page must have a main page title which is an H1, what do you propose instead?

carolinan:
On the single post template I can not get wide and full width blocks in the post content to show correctly after changing from full width to wide width.

I think the container needs to be kept at full width, and other means need to be used to constrain the post header, featured image, and post navigation.

jasmussen:
Nice work!

beafialho:
In my testing, I'm still seeing the featured images with original aspect ratio and the query loop full width. The other changes look good.

carolinan:
There is a problem with the single post pattern that I have not been able to solve.
When I use the photo blog single design, and I create a new post with default, wide, and full width content, the editor and front does not match.
In the editor, the full width group block expands outside the content area.
On the front, the full width group is the same width as the wide width.


https://github.com/user-attachments/assets/4066db38-026f-4ecc-9b2b-986480a15f33


(I also found it surprising that when the group is wide or full, the inner block is centered, but on the default width group it is left aligned)

carolinan:
The file names of the patterns use this format: template-single-photo.php.  
Would template-photo-single.php be better?

beafialho:
> There is a problem with the single post pattern that I have not been able to solve.
> When I use the photo blog single design, and I create a new post with default, wide, and full width content, the editor and front does not match.

I believe this may be because the template uses wide width, but when you're editing a post it doesn't reflect that constraint. I've seen this issue before.

> The file names of the patterns use this format: template-single-photo.php.
> Would template-photo-single.php be better?

What do you think about naming them more descriptively? Like, for example:

- Photo blog posts (photo-blog-posts.php)
- Photo blog single post (photo-blog-single-post.php)
- Photo blog archive (photo-blog-archive.php)
- Photo blog search results (photo-blog-search-results.php)

carolinan:
I think it is important that the file name includes 'template' to separate them from other types of patterns.

carolinan:
> I believe this may be because the template uses wide width, but when you're editing a post it doesn't reflect that constraint. I've seen this issue before.

Yeah, I don't like it but I think it can be considered an editor bug.

beafialho:
That makes sense 👍 Perhaps we can add "template" to the end of each?

`photo-blog-posts-template.php`

jasmussen:
> Yeah, I don't like it but I think it can be considered an editor bug.

It's likely I'm missing nuance, so apologies if this is off. But at one point, the post editor would show whatever content-width was applied to the content in the post template. What happened was that in layouts, e.g. two-column sidebar layouts, for the layout to work, content-width would not be applied and the post editor would show the content go edge to edge, which was a terrible writing experience, and not anywhere near the WYSIWYG mapping to the site editor that was intended. The decision was made at the time to rewind this and move forward again when we can actually load part of the posts template into the post editor.

patterns/photo-blog-home-template.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+?>
+
+<!-- wp:template-part {"slug":"header","area":"header","tagName":"header"} /-->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
+<main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
+	<!-- wp:heading {"textAlign":"center","level":1,"fontSize":"x-small"} -->
+	<h1 class="wp-block-heading has-text-align-center has-x-small-font-size">Stories</h1>
```
juanfra, line 21: I see that the pill/capsule style has been added in other PR. 

From the [figma discussion](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK?node-id=3177-1481#913568594) I understand that this element should be a paragraph, though. If that's correct we can chime in [here](https://github.com/WordPress/twentytwentyfive/pull/118/files#diff-af28145e38648fd038cea521ce238cdbb8b613e45e578cf8feff49774c066d1fR6) to mention that we need to include the paragraph block as well.

patterns/photo-blog-home-template.php
```diff
@@ -0,0 +1,32 @@
...
+	<!-- /wp:heading -->
+
+	<!-- wp:heading {"textAlign":"center","fontSize":"xx-large"} -->
+	<h2 class="wp-block-heading has-text-align-center has-xx-large-font-size">Tell your story</h2>
+	<!-- /wp:heading -->
+
+	<!-- wp:pattern {"slug":"twentytwentyfive/photo-blog-posts"} /-->
```
juanfra, line 28: I'm seeing this one take the full width of the screen. I believe it should take the wide width. Having a wrapper here could be the way to go, as it'd be for the template, and it'll make the photo blog posts pattern more flexible.

Screenshot 2024-08-28 at 15 38 54

patterns/photo-blog-single-post-template.php
```diff
@@ -0,0 +1,90 @@
...
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+<!-- wp:template-part {"slug":"header","area":"header","tagName":"header"} /-->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
+<main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
+	<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group alignwide" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)">
```
juanfra, line 20: I believe the design intends to have only the image as full width, but this part and the content should also be wrapped in site wide.

juanfra (commented):
Great work, Carolina! I left a few comments of things I found.

patterns/photo-blog-home-template.php
carolinan: Placing paragraphs above headings is not accessible. Having an H1 at this size is also not great.

beafialho: @richtabor do you have any helpful input regarding this?

richtabor: > Placing paragraphs above headings is not accessible. 

Placing a paragraph above a heading does not inherently make content inaccessible. The key to accessibility is ensuring that the headings are used in a logical, hierarchical order (e.g., H1, H2, H3) and that they accurately reflect the structure of the content.

richtabor: Have we thought about adding a block style variation for this though? Seems like it'd be a nice way to showcase as a "Text style".

patterns/photo-blog-home-template.php
carolinan: I am not able to reproduce this, can you write what steps you took to get to this result?
Are you sure there is no caching breaking any updates?

carolinan: It is not only the width of the section that is different, the spacing is also different than from what I am seeing.
I am using Chrome.

patterns/photo-blog-single-post-template.php
carolinan: I don't understand. It is set to alignwide.
```
<!-- wp:group {"align":"wide",
<div class="wp-block-group alignwide"
```

patterns/photo-blog-home-template.php
carolinan: This is inaccurate:
> Placing a paragraph above a heading does not inherently make content inaccessible. The key to accessibility is ensuring that the headings are used in a logical, hierarchical order (e.g., H1, H2, H3) and that they accurately reflect the structure of the content.

For example, if a category is shown above a post title, anyone who navigates the page using headings, will not receive the information about what category the post is in.

But I think this must be discussed in a separate issue, not here.

carolinan: > Have we thought about adding a block style variation for this though? Seems like it'd be a nice way to showcase as a "Text style".

Yes! it has only been a matter of time. The linked PR adds one.

patterns/photo-blog-home-template.php
juanfra: @carolinan here's what I did:

- Open the site editor.
- Go to the home blog template.
- Pick the new design from the design section.
- See the design applied and view it on the front end.

I'm using Firefox on a viewport bigger than `1440px`. I also checked Chrome, and I see the same results regarding spacing and the content taking the full width instead of the wide width. 

This could be similar to what we fixed [here](https://github.com/WordPress/twentytwentyfive/pull/129). 

What's the viewport size for you when viewing these templates? If it's `1440px` or less, it's likely that you won't see these problems.

patterns/photo-blog-single-post-template.php
juanfra: This can be pretty much related to what I [added here](https://github.com/WordPress/twentytwentyfive/pull/123#discussion_r1735829523) as well.

patterns/photo-blog-home-template.php
richtabor: then that solves this yea?

patterns/photo-blog-home-template.php
carolinan: I can see from the screenshot that the selected element, the UL for the post template, has `alignfull`. That is not part of the current PR.

Settings:
https://github.com/WordPress/twentytwentyfive/pull/123/files#diff-ecab6825b46c8296752891524bf1dd8316a64712982933886d988c9356b98210R27

The group block that is wrapping the post template has layout type default, so the align full option is not even available on the post template block.

Can you share the markup for the post template on the front and also double check that the in the editor, the post template is not set to full width because then the template has still somehow not been reset.

juanfra: Thanks for looking into this, Carolina! It's looking good to me now.

There must be something wrong with the caching; I had to do many things to see it (switching themes, resetting) finally. I did a template reset before leaving the previous comment and re-testing this PR, so apologies but I don't know what could have happened. 

I also have these in my `wp-config.php`. That should be enough to not have to deal with caching issues, right?

```
define( 'WP_ENVIRONMENT_TYPE', 'local' );
define( 'WP_DEVELOPMENT_MODE', 'theme' );
```

carolinan: I thought so but I have experienced the same and asked in core-editor on Slack.

carolinan:
> It's looking good to me now.

Based on this comment, I am going to merge this, adjustments can continue to be made after another design review.

PR #129, opened 2024-08-27
Fix headers and footers layout and spacing.

juanfra:
**Description**

According to the style's section on Figma, the [large screen behavior](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=77-7932&t=pAR7fcijMNjWyyje-1) is that the content (including header and footer) should be using the wide width as the max.

This PR adjusts the spacing of all headers and footers, so that they respect this. And it also adjusts the vertical padding of the default header to use 30px as the design.

**Screenshots**

https://github.com/user-attachments/assets/31c847d7-7cb1-473f-93c5-b9a21f5b6f9b

Screenshot 2024-08-27 at 11 03 38

**Testing Instructions**

1. Open the site editor.
2. Go to edit the header.
3. Select any of the designs included with TT5 and save.
4. Open the site on a large screen (more than 1400px wide)
5. Confirm that the headers are wrapped in the wide width defined in the design system.
6. Repeat with footers.
7. Ope

### DIFF
--- a/patterns/footer-centered.php
+++ b/patterns/footer-centered.php
@@ -12,15 +12,19 @@
  */
 
 ?>
-<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|50","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}}},"layout":{"type":"constrained"}} -->
-<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--70);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)">
-	<!-- wp:site-title {"level":0,"textAlign":"center"} /-->
-	<!-- wp:site-tagline {"textAlign":"center"} /-->
-	<!-- wp:spacer {"height":"20px"} -->
-	<div style="height:20px" aria-hidden="true" class="wp-block-spacer"></div>
-	<!-- /wp:spacer -->
-	<!-- wp:paragraph {"align":"center","fontSize":"x-small"} -->
-	<p class="has-text-align-center has-x-small-font-size">Designed with <strong>WordPress</strong></p>
-	<!-- /wp:paragraph -->
+<!-- wp:group {"layout":{"type":"constrained"}} -->
+<div class="wp-block-group">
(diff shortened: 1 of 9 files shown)


carolinan:
But that doesn't match any of the template designs, where the spacing between the edge and content is 50px:

https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=559-1863&m=dev

beafialho:
Thank you @juanfra! From your screen recording, it seems to work as intended.

The Figma templates show the desktop at 1440px. Beyond that, in larger screens, my intent was that the content should stop at 1440 px, which is what I tried to express in the "large screens behaviour". We should be mindful of how the patterns look within these alignments too.

juanfra:
> But that doesn't match any of the template designs, where the spacing between the edge and content is 50px

@carolinan, thanks! I assume you are talking about the vertical spacing of the header. I was calculating it differently. I used the final computed height of `90px` as a reference, which, in theory, matches the changes I initially added to the PR. Now I see the spacing of 20px in Figma. Although it makes the header `70px` height and not `90px`. I'll then revert the vertical spacing for the default header, unless we want to have a second look at it. Here's a quick screencast of what I mean.

https://github.com/user-attachments/assets/b0286f56-c2c6-4982-bb4c-21a1f79cc7d0

> The Figma templates show the desktop at 1440px.

@beafialho thanks! Right, the Figma templates show the desktop at a `1440px` width, with a left/right spacing of `50px` each. Which makes it `1340px` (wide width). These new changes take care of that. I'll be mindful about this when reviewing PRs for patterns, thank you.

carolinan:
No I was talking about the 50px horizontal spacing that is on all the default templates on Figma.
But as it was brought up, the different footers and headers do not use the same vertical spacing in Figma.

It would have been helpful to have had this reviewed while the original PR's for the headers, footers, and default templates were open.

carolinan:
Also the header height will depend on the content, for example when the menu expand to more than one row.

beafialho:
> It would have been helpful to have had this reviewed while the original PR's for the headers, footers, and default templates were open.

This one seems addressed. But for others perhaps we can hold on these adjustments when a full design review is done? I can test all on Playground and submit changes via a PR. Let me know what works best for you.

rejaulalomkhan:
@juanfra  Thanks for the test instructions. I tested the header and footer on a large screen `more than 1400px widescreen`. 
It's working for me.

juanfra:
> Also the header height will depend on the content, for example when the menu expand to more than one row.

Right, indeed. I was using the default styles of the default header contents as a reference. I reverted the vertical spacing changes on the default header to their previous state.

PR #140, opened 2024-08-27
Claim copyright for contributors.

peterwilsoncc:
**Description**

 One of the benefits of contributing to WordPress is that contributors license their code under the GPL rather than assign copyright to the WordPress Foundation or another entity.

As such, the bundled theme licenses ought to include a copyright claim for contributors similar to that in the main WordPress license.txt file.

**Screenshots**

N/A

**Testing Instructions**

N/A


juanfra (approved):
Thank you @peterwilsoncc - Looks good to me ✅

PR #141, opened 2024-08-28
Block style: Code 

### LINKED ISSUE #127: Block style: Code

Add the default style for the code block to theme.json.
The background color should be `secondary`, and the text color should be `contrast`.
Font size should be `small`. Font weight `400`. The font family should not be set, the block can fallback to the default. 
The left, right, top and bottom padding should be `small`.

[Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=565-10056&m=dev)

(Note that the design in Figma uses other color names)

kishanjasani:
**Description**
Update the styles for the default code block. Fixes: #127 

**Screenshots**
Screenshot at Aug 28 09-22-53

**Testing Instructions**

1. Add page
2. Add code block to the page and check the styles.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -953,7 +953,6 @@
 				"color": {
 					"background": "var:preset|color|secondary",
 					"text": "var:preset|color|contrast"
-
 				},
 				"spacing": {
 					"padding": {
--- a/theme.json
+++ b/theme.json
@@ -947,8 +947,21 @@
 		"blocks": {
 			"core/code": {
 				"typography": {
-					"fontFamily": "monospace",
-					"fontSize": "var:preset|font-size|small"
+					"fontSize": "var:preset|font-size|small",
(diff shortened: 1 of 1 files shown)


juanfra (commented):
Thanks for working on this. The changes look good ✨

The only thing I noticed is that the figma is using the `SF Mono` font and given that we're not including it, the code block is defaulting to `monospace`. 

@beafialho are we ok using `monospace` here?

beafialho:
We could perhaps use `Fira Code`, since we're including it in the theme?

juanfra:
Thanks Bea! Sounds good to me. @kishanjasani can you please work on the adjustments?

kishanjasani:
Updated the fonts to `fira-code`.

PR #144, opened 2024-08-28
Personal blog: Misc width changes

carolinan:
**Description**
After the change to the width of the header and footer, some elements in the default templates have the wrong width.

**Screenshots**
Query pagination before:
pagination-before

Query pagination after:
pagination-after

Post navigation (next and previous post link) before:
post-nav-before

Post navigation (next and previous post link) after:
post-nav-after

( I had to use the zoom on the browser to take these shots, I hope that does not distort them too much)

**Testing Instructions**
On your WordPress test installation, create a new post with a default width group block, a wide, and full width group block.
Make sure that your test installation has enough posts to test the query pagination and the next and previous posts navigation.
In the admin area, set the reading setting to show the latest posts on the home page.

On the front: View the homepage and confirm that default, wide, and full width blocks display correctly.
Confirm that the query pagination aligns with the wide width, the site header, and site footer.

On the front: Go to the post you created that has the blocks with the different widths.
Confirm that the next and previous posts links align with the wide width, site header, and site footer.

### DIFF
--- a/patterns/post-navigation.php
+++ b/patterns/post-navigation.php
@@ -12,10 +12,10 @@
  */
 
 ?>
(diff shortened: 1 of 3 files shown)


beafialho:
Thanks @carolinan these look good to me!

Issue #142, opened 2024-08-28
Add section styles


carolinan:
This is a **suggestion** for adding section styles, block style variations that can be applied to, in this case, the group block.

Each style would use an accessible combination of two or more colors from the palette, **and each style needs to continue to work when the user changes the color palette preset.** 

The purpose is to make it easy for users to change the style of a part of their content, pattern, or templates.

Section styles are .JSON files placed in the `styles` folder or a subfolder to `styles`.
Each file needs to include: schema, version, title, slug, one ore more block types, and the styles.

Section styles can not include settings. For example, it can set a font size, but not add new font size options to the typography control.

```
{
	"$schema": "https://schemas.wp.org/trunk/theme.json",
	"version": 3,
	"title": "Base",
	"slug": "section-1",
	"blockTypes": [ "core/group" ],
	"styles": {}
}
```

Each the section style would include:

- Background color and text color. Link color, if different from text color.
- Spacing
- Styles for elements:
  - Buttons
  - Headings

The CSS for these style variations are often repeated multiple times on the same page, which is why I recommend limiting the styles.

richtabor:
Here's an example I've found pretty solid (we'd replace colors, make sure we do what's suitable for each): 

```
{
    "$schema": "https://schemas.wp.org/trunk/theme.json",
    "version": 3,
    "slug": "section-3",
    "title": "Style 3",
    "blockTypes": [
        "core/group",
        "core/columns",
        "core/column"
    ],
    "styles": {
        "color": {
            "background": "var(--wp--preset--color--theme-5)",
            "text": "var(--wp--preset--color--theme-1)"
        },
        "blocks": {
            "core/separator": {
                "color": {
                    "text": "color-mix(in srgb, currentColor 25%, transparent)"
                }
            }
        },
        "elements": {
            "button": {
                "color": {
                    "background": "var(--wp--preset--color--theme-1)",
                    "text": "var(--wp--preset--color--theme-5)"
                },
                ":hover": {
                    "color": {
                        "background": "color-mix(in srgb, var(--wp--preset--color--theme-1) 85%, #000)",
                        "text": "var(--wp--preset--color--theme-5)"
                    }
                }
            }
        }
    }
}
```

richtabor:
Worth nothing that each color variation should define its own as well, like this: https://github.com/Automattic/themes/blob/trunk/assembler/styles/colors/01-blueberry.json#L155-L202

jasmussen:
[...]

PR #151, opened 2024-08-29
Try: Add 3 section styles, update Hero: Centered Heading pattern.

jasmussen:
**Description**

Starts initial work on adding Section styles (#142). In this case, it adds three section styles from [this spot in the Figma file](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=77-258&t=0tJE2VXhthA9HmZK-1):

Screenshot 2024-08-29 at 11 53 00

These use what I believe is the [most suggested format for section style slugs](https://github.com/WordPress/twentytwentyfive/issues/3#issuecomment-2200847735) (#3), `section-1`, `section-2`, `section-3`.

Additionally, I made a change to the Hero Centered Heading pattern to remove the explicitly set heading color, so that it will inherit the color set on the container instead.

**Screenshots**

section styles

**Testing Instructions**

1. Open the inserter, insert 4 instances of the Hero Centered Heading pattern.
2. Set the 2nd to fourth pattern to styles 2-4 in the inspector.
3. Test additional patterns (group/column based) for these section styles.

### DIFF
--- a/patterns/hero-centered-heading.php
+++ b/patterns/hero-centered-heading.php
@@ -15,8 +15,8 @@
 <div class="wp-block-group alignfull" style="min-height:0vh;margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--70);padding-right:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--70);padding-left:var(--wp--preset--spacing--40);">
 	<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|40"}},"layout":{"type":"constrained"}} -->
 	<div class="wp-block-group">
-		<!-- wp:heading {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"right":"0","left":"0"},"padding":{"right":"0","left":"0"}},"elements":{"link":{"color":{"text":"var:preset|color|contrast"}}}},"textColor":"contrast"} -->
-		<h1 class="wp-block-heading has-text-align-center has-contrast-color has-text-color has-link-color" style="margin-right:0;margin-left:0;padding-right:0;padding-left:0">Tell your story</h1>
+		<!-- wp:heading {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"right":"0","left":"0"},"padding":{"right":"0","left":"0"}}}} -->
+		<h1 class="wp-block-heading has-text-align-center" style="margin-right:0;margin-left:0;padding-right:0;padding-left:0">Tell your story</h1>
 		<!-- /wp:heading -->
 
 		<!-- wp:paragraph {"align":"center"} -->
--- a/styles/blocks/section-1.json
+++ b/styles/blocks/section-1.json
@@ -26,6 +26,11 @@
 					"background": "var(--wp--preset--color--contrast)",
 					"text": "var(--wp--preset--color--base)"
 				}
+			},
+			"link": {
+				"color": {
+					"text": "var(--wp--preset--color--contrast)"
+				}
 			}
 		}
 	}
--- a/styles/blocks/section-1.json
+++ b/styles/blocks/section-1.json
@@ -1,0 +1,32 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"slug": "section-1",
+	"title": "Style 1",
+	"blockTypes": [
+		"core/group",
+		"core/columns",
+		"core/column"
+	],
+	"styles": {
+		"color": {
+			"background": "var(--wp--preset--color--accent-2)",
+			"text": "var(--wp--preset--color--contrast)"
+		},
+		"blocks": {
+			"core/separator": {
+				"color": {
+					"text": "color-mix(in srgb, currentColor 25%, transparent)"
+				}
+			}
+		},
+		"elements": {
+			"button": {
+				"color": {
+					"background": "var(--wp--preset--color--contrast)",
+					"text": "var(--wp--preset--color--base)"
+				}
+			}
+		}
+	}
+}
--- a/styles/blocks/section-2.json
+++ b/styles/blocks/section-2.json
@@ -1,0 +1,32 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"slug": "section-2",
+	"title": "Style 2",
+	"blockTypes": [
+		"core/group",
+		"core/columns",
+		"core/column"
+	],
+	"styles": {
+		"color": {
+			"background": "var(--wp--preset--color--accent-3)",
+			"text": "var(--wp--preset--color--accent-2)"
+		},
+		"blocks": {
+			"core/separator": {
+				"color": {
+					"text": "color-mix(in srgb, currentColor 25%, transparent)"
+				}
+			}
+		},
+		"elements": {
+			"button": {
+				"color": {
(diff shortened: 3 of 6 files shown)


patterns/hero-centered-heading.php
```diff
@@ -15,8 +15,8 @@
 <div class="wp-block-group alignfull" style="min-height:0vh;margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--70);padding-right:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--70);padding-left:var(--wp--preset--spacing--40);">
 	<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|40"}},"layout":{"type":"constrained"}} -->
 	<div class="wp-block-group">
-		<!-- wp:heading {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"right":"0","left":"0"},"padding":{"right":"0","left":"0"}},"elements":{"link":{"color":{"text":"var:preset|color|contrast"}}}},"textColor":"contrast"} -->
-		<h1 class="wp-block-heading has-text-align-center has-contrast-color has-text-color has-link-color" style="margin-right:0;margin-left:0;padding-right:0;padding-left:0">Tell your story</h1>
+		<!-- wp:heading {"textAlign":"center","level":1,"style":{"spacing":{"margin":{"right":"0","left":"0"},"padding":{"right":"0","left":"0"}}}} -->
```
jasmussen, line 18: @juanfra let me know if it was intentional to explicitly set the heading color here.

styles/blocks/section-1.json
```diff
@@ -0,0 +1,32 @@
...
+	],
+	"styles": {
+		"color": {
+			"background": "var(--wp--preset--color--accent-2)",
+			"text": "var(--wp--preset--color--contrast)"
+		},
+		"blocks": {
+			"core/separator": {
+				"color": {
+					"text": "color-mix(in srgb, currentColor 25%, transparent)"
```
jasmussen, line 19: @carolinan @beafialho Let me know if I got this right! I only found a few separator designs in the Figma.

patterns/hero-centered-heading.php
juanfra: No, it was not intentional.

richtabor:
> Additionally, I made a change to the Hero Centered Heading pattern to remove the explicitly set heading color, so that it will inherit the color set on the container instead.

We should remove explicit color application almost everywhere.

jasmussen:
> We should remove explicit color application almost everywhere.

I found another instance here:
Screenshot 2024-08-29 at 14 44 19

and pushed a fix to this:

Screenshot 2024-08-29 at 14 45 46

I'm not going to push more to this branch, just to keep it focused, but I'll spin up a separate PR to fix an issue with Large Title:
Screenshot 2024-08-29 at 14 52 41

Also separate, but related question, any advice on what to do with patterns that really look best with color? Example:

color

A few things at play here:

* That pattern really looks most compelling in the pattern browser when it has that base yellow background color. Removing it loses a lot of character. Can we default a pattern to a _Section style_?  Guess we could, right?
* Any idea why the link color isn't inherited? I guess that should be defined in these section styles too, yes?

jasmussen:
Actually I got that yellow pattern working pretty well. Okay to push to this branch too?

working

jasmussen:
If no-one minds, I'd like to push the pattern update, it's working quite well, and it's a great way to provide some sectional diversity in the bundled patterns:

pattern with preset section style

carolinan:
> That pattern really looks most compelling in the pattern browser when it has that base yellow background color. Removing it loses a lot of character. Can we default a pattern to a Section style? Guess we could, right?

Yes, that was the plan and why the color wasn't added earlier.

richtabor:
> Also separate, but related question, any advice on what to do with patterns that really look best with color? 
> Yes, that was the plan and why the color wasn't added earlier.

The background color should not be applied, but a section style can be added as a class to the pattern (any pattern).

styles/blocks/section-1.json
```diff
@@ -0,0 +1,37 @@
...
+		"elements": {
+			"button": {
+				"color": {
+					"background": "var(--wp--preset--color--contrast)",
+					"text": "var(--wp--preset--color--base)"
+				}
+			},
+			"link": {
+				"color": {
+					"text": "var(--wp--preset--color--contrast)"
```
richtabor, line 32: I think we could remove this, if we used currentColor in the parent theme.json.

richtabor: Otherwise if the user changes the text color, they'll also need to also change the link color.

jasmussen:
> The background color should not be applied, but a section style can be added as a class to the pattern (any pattern).

Nice, yeah that's what I did in [this commit](https://github.com/WordPress/twentytwentyfive/pull/151/commits/0844f5079e56c24a23073481b66183f4f3d8f9df). Works well.

styles/blocks/section-2.json
```diff
@@ -0,0 +1,37 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"slug": "section-2",
```
richtabor, line 4: Food for thought, `is-style-section-2` will be the class name here. Just thinking, should they be `is-stye-pattern-2`—as these section styles are generally applied at the "pattern" level?

styles/blocks/section-1.json
jasmussen: Oh yeah, good call!

juanfra (approved):
This is so good! Works well for me, and code lalso ooks good.

https://github.com/user-attachments/assets/264e223f-5adf-4cc2-ba31-aba3a1cd4011

styles/blocks/section-2.json
juanfra: My two cents: Although they're going to be generally applied at the pattern level, it _could not_ be 100% of the cases, and for those edge cases, it can be confusing to have something named pattern when it's not.

At a theme development level, it will make sense. But a site admin can create a group of blocks and apply styles without the whole thing being a pattern. I believe that section is more generic.

justintadlock: This goes back to this discussion: https://github.com/WordPress/twentytwentyfive/issues/3

I would avoid tying the slug to patterns just because that's where they're generally being used in this theme. Section/block styles can be used at any level (on the pattern wrapper, in sub-sections of patterns, in templates, just on any ol' block that it's registered for). Using "pattern" in the slug just muddies the waters a bit and will limit the use cases that folks imagine. Keep it to the more generic `section` prefix and be done with it.

jasmussen:
Thanks all for the reviews and input. I'm going to let this one sit unmerged until tomorrow morning, to let [this conversation settle](https://github.com/WordPress/twentytwentyfive/pull/151#discussion_r1736282184), and I'll implement what gets decided. It won't necessarily be set in stone from that point onwards, we'll always be able to change things until the beta period, but it will be good to land this soon, so all new patterns created can use best practices as far as how to apply colors!

styles/blocks/section-2.json
richtabor: Alrighty, just asking to see what you think. Thanks!

jasmussen:
Coming back this morning, it seems like the conversation ended on the same spot that this PR landed on. So I'm going to go ahead and merge, knowing that this is still possible to change after this merge, I'll be happy to address any followups. Thanks all!

PR #153, opened 2024-08-29
Try: Reduce spacing presets to use 1-click controls

### LINKED ISSUE #152: Spacing/Typography presets: reduce total amount to use the ergonomic controls

At the moment, there are 8 spacing presets, and 6 typography presets. These amounts cause the segmented dimensions and typography controls to be replaced by dropdowns:

Screenshot 2024-08-29 at 12 19 32

Screenshot 2024-08-29 at 12 22 05

**Describe the solution you'd like**

The dropdown experience is not as good a user experience, requiring two clicks instead of one. If we reduce the amount of spacing presets to 7 (for example deleting the X-Tiny), we get the range control:

Screenshot 2024-08-29 at 12 24 43

Similarly, if we remove a single font size preset, we get the single-click typography size control:

Screenshot 2024-08-29 at 12 28 52

I'd suggest consolidating and removing either the Extra Large or Extra Extra Large size.

jasmussen:
**Description**

Fixes #152. Eliminates one preset from both spacing and typography preset arrays, so that they use the 1-click controls rather than the extended dropdown controls:

Screenshot 2024-08-29 at 12 36 13

Specifically this removes the X-Tiny (5px) preset, and it removes the X-Large preset and changes the XX-Large preset to take its place. 

That's a test to see how it feels out, please take a look and I'll update the PR accordingly.

**Testing Instructions**

1. Open a page or post
2. Type some text, observe the inspector that it has the 1-click segmented control font size picker.
3. Add a padding control from the ToolsPanel, observe that it has the 1-click range control for spacing presets.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -55,11 +55,6 @@
 			"defaultSpacingSizes": false,
 			"spacingSizes": [
 				{
-					"name": "X-Tiny",
-					"size": "5px",
-					"slug": "10"
-				},
-				{
 					"name": "Tiny",
 					"size": "10px",
 					"slug": "20"
@@ -143,21 +138,12 @@
 				},
 				{
 					"fluid": {
-						"max": "3rem",
-						"min": "2.375rem"
-					},
-					"name": "Extra Large",
-					"size": "3rem",
-					"slug": "x-large"
-				},
-				{
-					"fluid": {
 						"max": "3.625rem",
 						"min": "2.625rem"
 					},
-					"name": "Extra Extra Large",
+					"name": "Extra Large",
 					"size": "2.625rem",
-					"slug": "xx-large"
+					"slug": "x-large"
 				}
 			],
 			"fontFamilies": [


juanfra (approved):
Code looks good to me 🚀

I've only found one appearance of the x-tiny spacing preset [here](https://github.com/WordPress/twentytwentyfive/blob/trunk/templates/single.html#L11), which if I'm not mistaken is for [this](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3326-1733&t=iVc8k2Kl9C1qrITw-1). I believe we can get rid of this spacing, hardcode it or set it to tiny.

carolinan:
If we need any more sizes I recommend using the `custom `setting in theme.json to create them.

jasmussen:
Great feedback both. Is this good to land as-is? I'll see if I can follow up on the other items when I get a moment. Thank you for the reviews 🙏

jasmussen:
Followed up in #154 on the spacing preset. Thanks again.

PR #156, opened 2024-08-29
Try: Add comments.

### LINKED ISSUE #115: Add comments area to posts

- [x] The theme needs a design for the comments area (including comment pagination and the form).
- [ ] the design also needs to be implemented as a pattern.

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/115

- Adding comments configuration to `theme.json`
- Adding comments hidden pattern.
- Including comments in the single template.

🗒️ Notes

1. I set some `5px` values now that we no longer have that as a preset.
2. I had to set some custom CSS to override the default styles of the form inputs.

**Screenshots**

https://github.com/user-attachments/assets/4609b696-7e14-4fe7-8518-0af52a333667

**Testing Instructions**

1. Create a post, be sure that the template is using the comments.
3. Leave as many comments as you like.
4. Confirm that the implementation looks like [the designs](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3326-2103&t=iVc8k2Kl9C1qrITw-1)

### DIFF
--- a/patterns/hidden-comments.php
+++ b/patterns/hidden-comments.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Title: Comments
+ * Slug: twentytwentyfive/hidden-comments
+ * Inserter: no
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+<!-- wp:comments {"className":"wp-block-comments-query-loop","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} -->
+<div class="wp-block-comments wp-block-comments-query-loop" style="margin-top:var(--wp--preset--spacing--70);margin-bottom:var(--wp--preset--spacing--70)">
+	<!-- wp:heading -->
+	<h4><?php esc_html_e( 'Comments', 'twentytwentyfive' ); ?></h4>
+	<!-- /wp:heading -->
+	<!-- wp:comments-title {"level":4} /-->
+	<!-- wp:comment-template -->
+	<!-- wp:group {"style":{"spacing":{"margin":{"top":"0","bottom":"var:preset|spacing|50"}}}} -->
+	<div class="wp-block-group" style="margin-top:0;margin-bottom:var(--wp--preset--spacing--50)">
+		<!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap","verticalAlignment":"top"}} -->
+		<div class="wp-block-group">
+			<!-- wp:avatar {"size":50} /-->
+			<!-- wp:group -->
+			<div class="wp-block-group">
+				<!-- wp:comment-date /-->
+				<!-- wp:comment-author-name /-->
+				<!-- wp:comment-content /-->
+				<!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap"}} -->
+				<div class="wp-block-group">
+					<!-- wp:comment-edit-link /-->
+					<!-- wp:comment-reply-link /-->
+				</div>
+				<!-- /wp:group -->
+			</div>
+			<!-- /wp:group -->
+		</div>
+		<!-- /wp:group -->
+	</div>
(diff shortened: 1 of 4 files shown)


patterns/hidden-comments.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
jasmussen: I wish twe didn't have to store templates in the patterns folder in order for them to be translatable. I guess there's no real way around that?

theme.json
```diff
@@ -931,6 +931,11 @@
 			"lineHeight": "1.4"
 		},
 		"blocks": {
+			"core/avatar": {
+				"border": {
+					"radius": "100px"
```
jasmussen, line 936: Would it make sense to create some radius presets, of which one preset could be "round", and just set to 9999px?

theme.json
```diff
@@ -943,33 +948,131 @@
...
+							"text": "var:preset|color|contrast"
+						}
+					}
+				},
 				"typography": {
 					"fontSize": "var:preset|font-size|small"
 				}
 			},
+			"core/post-comments-form": {
+				"css": "& textarea, input:not([type=submit]){border-radius:.25rem; border-color: var(--wp--preset--color--opacity-20) !important;} & input[type=checkbox]{margin:0 .2rem 0 0 !important;} & label {font-size: var(--wp--preset--font-size--x-small); }",
```
jasmussen: Some finnicky details here. But notably the input radius It'd be nice to again add to a preset scale, so they're consistent across.

jasmussen (commented):
PR looks good to me, works well. I left a few comments. Short of approving this, as I'm feeling out how I can best contribute to this project, I'll defer to Carolina and/or Bea on details. Though it seems like we could merge and iterate.

patterns/hidden-comments.php
carolinan: No. I don't know if there is an open issue that covers it better than this one:
https://github.com/WordPress/gutenberg/issues/20966

The project is closer to solving it now that there are block bindings.
_In theory,_ block bindings could already be used, since you can use custom PHP functions which could include the translations.... but I don't think this theme should experiment with it.

theme.json
carolinan: There is a request in the Gutenberg repository to add this to the editor.
Let's check how far that has come, but I think it needs dev:
https://github.com/WordPress/gutenberg/issues/64041

carolinan: I will add the issue to the list of issues.

carolinan:
Because the page templates are designed to not have comment areas _by default_, I think it would be good to not make this a hidden pattern. That way the user can add a new comment area that looks as good.

beafialho:
Thank you, this is looking good to me!

patterns/hidden-comments.php
```diff
@@ -0,0 +1,51 @@
...
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+<!-- wp:comments {"className":"wp-block-comments-query-loop","style":{"spacing":{"margin":{"top":"var:preset|spacing|70","bottom":"var:preset|spacing|70"}}}} -->
+<div class="wp-block-comments wp-block-comments-query-loop" style="margin-top:var(--wp--preset--spacing--70);margin-bottom:var(--wp--preset--spacing--70)">
+	<!-- wp:heading -->
+	<h4><?php esc_html_e( 'Comments', 'twentytwentyfive' ); ?></h4>
```
carolinan: This heading is missing the block class name and is creating a block validation error.

carolinan: Why is the H4 heading used twice? The hierarchy is incorrect.
Update: "Comments" should be a H2 and the comments title an H3. The comment reply link is an H3 that can not be changed.

patterns/hidden-comments.php
carolinan: Also, this particular part of the theme is not a template or a part, it is intended to be a pattern so that it can be inserted anywhere, and it would be PHP file even if it had no translatable strings.

PR #157, opened 2024-08-29
Try: Add buttons styles.

juanfra:
**Description**

Adding button styles according to [the designs](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=2431-1113&t=iVc8k2Kl9C1qrITw-1).

It works well in the default style. We will probably need to revisit the different variations once we have the [section styles](https://github.com/WordPress/twentytwentyfive/pull/151) in.

**Screenshots**

https://github.com/user-attachments/assets/86fb4301-931b-4d78-bed2-17b1aeb4299e

**Testing Instructions**

1. Create a page.
2. Add a button with the default style and another one with outline.
3. Confirm they're looking like [the designs](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=2431-1113&t=iVc8k2Kl9C1qrITw-1).


assets/css/button-outline.css
```diff
@@ -0,0 +1,6 @@
+.wp-block-button.is-style-outline
```
jasmussen, line 1: If there was a way to avoid this custom stylesheet, it would be a win. What is the downside to removing this piece?

juanfra: I believe there's no way to set the outline variation hover styles in the `.json` file. Removing this file will cause the outline variation buttons not to have any styling on interactions.

jasmussen: I'd personally prefer that over adding custom CSS. The problem with custom CSS is that it isn't user-styleable. 

https://github.com/WordPress/gutenberg/issues/38277 may not land in time for 6.7, but TT5 will last a whole year, so if this UI becomes available in 6.8, we'll have a bit of technical debt.

juanfra: It's a tough one because, as a site owner, I'll feel the buttons are broken, and as a site visitor, I'd feel they're not available. I share your concern about custom CSS, I thought we could do what was done in TT4 and I was unaware of the GitHub issue you've shared. That will pretty much fix things, if it's also working with variations.

@carolinan thoughts?

jasmussen: CC: @richtabor as well, I think you may have had similar thoughts as relating to section styles, etc.

carolinan: Controversial? but I would not be opposed to not supporting the outline button. 
I mean deregistering the style, in the theme.
To me, this is a legacy style much like the older quote block or pull quote styles, because today the user can create the same style with the block options.

The outline button creates some different styling problems, for example if you style the regular button, it can be a bit difficult to make the two buttons the same size without custom css.

carolinan: > I believe there's no way to set the outline variation hover styles in the .json file. Removing this file will cause the outline variation buttons not to have any styling on interactions.

Lets confirm if this still the case with all the recent CSS changes in the past releases?

beafialho:
Design-wise this is looking great to me 👍

Happy to see both styles have the same height:

Captura de ecrã 2024-08-30, às 16 07 56

assets/css/button-outline.css
juanfra: > Lets confirm if this still the case with all the recent CSS changes in the past releases?

I've been trying different things and they're not working. It looks like it was already reported [here](https://github.com/WordPress/gutenberg/issues/59731).

On a different note and unrelated, `outline` is a reserved word for styles in the `json`, which generates confusion and probably not the best devex if the variation is using the same name.

carolinan: That issue is from March though, the style variations were updated in June.
I think this is the most relevant issue and context: 
https://github.com/WordPress/gutenberg/issues/62686
https://github.com/WordPress/gutenberg/issues/64501

juanfra: Right, I've shared that one in particular because it described the same issue we have with the `:hover` state of the button for the outline variation.

I've tried the different ways of defining the variation for buttons and I defining styles for `:hover` is not working for me.

PR #163, opened 2024-08-30
Add two new section styles, and 

jasmussen:
**Description**

Adds two new section styles, fixes a pattern typo, and updates the Clients Section pattern to use section styles.

The two new styles are inspired by these styles in the Figma:

Screenshot 2024-08-30 at 13 01 34

Screenshot 2024-08-30 at 13 01 44

Screenshot 2024-08-30 at 13 01 48

Here's the end results:

state

Separately, an issue has surfaced with the Social Icons block:

icon colors

Basically, since we can't globally style the _icon color_, neither can we section-style the icon color. @getdave if you'll forgive the ping. In case you have time to look, do you know how big of a lift it would be in the editor to add global styles support for the Icon Color?

**Testing Instructions**

Try some patterns, including the "Clients section" pattern, with section styles. Test the new beige and black styles.

### DIFF
--- a/patterns/clients-section.php
+++ b/patterns/clients-section.php
@@ -11,8 +11,8 @@
  */
 
 ?>
-<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|80","bottom":"var:preset|spacing|80","left":"var:preset|spacing|50","right":"var:preset|spacing|50"},"margin":{"top":"0","bottom":"0"}}},"backgroundColor":"secondary","layout":{"type":"constrained"}} -->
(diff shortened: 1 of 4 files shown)


jasmussen:
Would it be possible/reasonable/feasible to, in theme.json, create a new color-swatch that maps to `currentColor` and apply it explicitly for the icon color?

carolinan:
> Would it be possible/reasonable/feasible to, in theme.json, create a new color-swatch that maps to currentColor and apply it explicitly for the icon color?

I have not tried if it works, or what custom CSS that would be needed for adding the color to the icon. 
Or are we expecting currentColor to work in the color **control** too?

beafialho:
This is looking good in my testing, the only issue I saw was the social icons color. Excited that it's working nicely!

beafialho (approved):

jasmussen:
> Or are we expecting currentColor to work in the color control too?

I think @richtabor suggested that _refs_ is the future forward way to solve this, which I'd tend to agree with, so a swatch that maps to currentColor is mostly theoretical at this point. It works, though, but as should also be clear: it's not a good user experience:

currentColor

PR #168, opened 2024-09-01
Pattern: More posts ( Previously "other posts")

### LINKED ISSUE #134: Question: How to address "Other Posts" section in the templates?
**Description**
As per [the design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=559-1893&t=bCFceo6KG1JJBgYQ-4), we have an "Other Posts" section in some of the Single templates. It seems that it isn't possible to style the latest posts block this way and the problem with using the query loop is that it can not exclude the current post, so it won't technically be "other" posts.

@richtabor @MaggieCabrera @jasmussen do you have a suggestion for how to build this? Should we reconsider using this section?

Captura de ecrã 2024-08-27, às 16 26 15

carolinan:
**Description**
Adds a new pattern with a list of posts with the post title and post date, below a heading with the text "More posts",
intended to show above the site footer in several templates.

Closes https://github.com/WordPress/twentytwentyfive/issues/134
[Figma link](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3326-1741&m=dev)

**Testing Instructions**

Apply the PR and view the updated templates. Confirm that the pattern matches the design.
- Home
- Index
- Archive
- Search
- Single post

To test the index, first rename or delete the home template, then the index template file will be used as the front page on the front of the site.

**Screenshots**

### DIFF
--- a/patterns/more-posts.php
+++ b/patterns/more-posts.php
@@ -1,0 +1,34 @@
+<?php
+/**
+ * Title: More posts
+ * Slug: twentytwentyfive/more-posts
+ * Description: Displays a list of posts with title and date.
+ * Categories: query
+ * Block Types: core/query
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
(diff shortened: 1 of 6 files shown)


beafialho:
Thank you @carolinan! This is looking great to me.

PR #169, opened 2024-09-01
Removed Empty line & also remove queryId in several files

shail-mehta:
#### Removed Empty line in below files: 
- patterns/news-blog-archive-template.php
- patterns/photo-blog-archive-template.php
- patterns/photo-blog-home-template.php
- patterns/photo-blog-search-results-template.php

#### Removed `queryId` in below files: 
- patterns/hidden-sidebar.php
- patterns/news-blog-home-template.php

### DIFF
--- a/patterns/hidden-sidebar.php
+++ b/patterns/hidden-sidebar.php
@@ -19,7 +19,7 @@
 <div style="height:var(--wp--preset--spacing--40)" aria-hidden="true" class="wp-block-spacer"></div>
 <!-- /wp:spacer -->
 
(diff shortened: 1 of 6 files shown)


carolinan:
Because this markup will change many times until it is finalised, is there any way I can motivate you to work on more useful tasks?

carolinan:
For example, if the query loop is updated in the editor, the id will be copy pasted back in, and we will end up removing it over and over. So these tasks are best left to the very end.

shail-mehta:
I completely agree with this point. It’s better to address this task later.

shail-mehta:
> Because this markup will change many times until it is finalised, is there any way I can motivate you to work on more useful tasks?

Yes, I will work on another pattern.

Issue #130, opened 2024-08-27
Footers: Adjustments to match the design


beafialho:
I tested the footers and made some modifications accordingly. Pasted the code to each below and before/after pictures.

### Default footer

| Before      | After |
| ----------- | ----------- |
|  ![Captura de ecrã 2024-08-27, às 11 59 57](https://github.com/user-attachments/assets/1dac0d88-7ec6-4cfb-9e27-115957ca0de8)   |  ![Captura de ecrã 2024-08-27, às 12 04 25](https://github.com/user-attachments/assets/0462545f-fce6-41a3-bcb2-3f4e7f9f5a16)   |

```
<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|50","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}}},"layout":{"type":"constrained"}} -->
<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)"><!-- wp:group {"align":"wide","layout":{"type":"default"}} -->
<div class="wp-block-group alignwide"><!-- wp:site-logo /-->

<!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between","verticalAlignment":"top"}} -->
<div class="wp-block-group alignfull"><!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|20","padding":{"top":"0","bottom":"0","left":"0","right":"0"}},"layout":{"selfStretch":"fit","flexSize":null}},"layout":{"type":"constrained"}} -->
<div class="wp-block-group" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0"><!-- wp:site-title {"level":2} /-->

<!-- wp:site-tagline /-->

<!-- wp:spacer {"height":"var:preset|spacing|40","width":"0px","style":{"layout":{}}} -->
<div style="height:var(--wp--preset--spacing--40);width:0px" aria-hidden="true" class="wp-block-spacer"></div>
<!-- /wp:spacer --></div>
<!-- /wp:group -->

[...]

PR #172, opened 2024-09-02
Update the markup of the footers

carolinan:
**Description**
Updates the markup of the footers.

I have used the markup from #130, with these exceptions:
- I re-added the aria labels to the navigation blocks
- I replaced the x-small font size with small, since x-small has been removed.
- I replaced the 20px spacers with the same size, but using the spacing preset.
- I replaced two background colors by applying the section styles, and removed hard coded colors.

Newsletter footer: Replace the site title with a heading with a single word
Footer with columns: Make it wide inside a full width group.

Closes
https://github.com/WordPress/twentytwentyfive/issues/130

**Testing Instructions**

Apply the PR
View the footers in the editor and the front, on different screen widths.

**Screenshots**


jasmussen:
Looks good. A few things that may or may not be part of this PR, the registered trademark symbol makes this word wrap:

Screenshot 2024-09-03 at 10 07 23

Regardless of what we may or may not replace that with, we probably need a slightly smaller font size there.

Bea mentioned these should be XL. They are XL in trunk, but appear to be Large here. 

Screenshot 2024-09-03 at 10 07 38

There's something going on with the font sizes, especially for the navigation, in this one:

Screenshot 2024-09-03 at 10 08 01

carolinan:
>There's something going on with the font sizes, especially for the navigation, in this one

I am not able to reproduce this. The heading above, and the navigation are set to use the medium size.
Individual custom links can have their font size changed on the block, but by default new items added to the navigation should all be medium.


----
I found that when one navigation block has more items, they don't align. In the wrapping row block, the vertical alignment is set to center. It should probably be aligned to top, what do you think?
image

beafialho:
> There's something going on with the font sizes, especially for the navigation, in this one:

@jasmussen I suspect the screenshot you shared is from a pattern from the pattern directory. The ones from TT5 are looking good to me.

beafialho:
Thank you @carolinan, I left a comment above but other than that, they're looking good on my end.

patterns/footer-columns.php
```diff
@@ -12,26 +12,27 @@
...
+<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--50);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50)">
+	<!-- wp:group {"align":"wide","layout":{"type":"default"}} -->
+	<div class="wp-block-group alignwide">
 		<!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between","verticalAlignment":"top"}} -->
 		<div class="wp-block-group alignfull">
 			<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|20","padding":{"top":"0","bottom":"0","left":"0","right":"0"}}},"layout":{"type":"constrained"}} -->
 			<div class="wp-block-group" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0">
-				<!-- wp:site-title {"level":2,"fontSize":"xx-large"} /-->
+				<!-- wp:site-title {"style":{"typography":{"fontSize":"70px"}}} /-->
```
carolinan: This must not be an H1, which is the default level for the site title.
And the custom CSS variable should be used for the 70px.

carolinan:
> Thank you @carolinan, I left a comment above but other than that, they're looking good on my end.

I'm sorry, which comment, the one about the core block?

beafialho:
> > Thank you @carolinan, I left a comment above but other than that, they're looking good on my end.
> 
> I'm sorry, which comment, the one about the core block?

This one:

For this one big heading, let's make a small change to better match the design:

Font size: **26vw**
Letterspacing: **-0.02em**
Appearance: **Semi Bold**

juanfra (commented):
It is working well to me, using the styles that are coming from the GH issue. 

I noticed the font weight on the "Stories" text for the "Footer with newsletter signup", but it looks like Bea already highlighted that.

carolinan:
26vw looks like this on my screen at 1920x1080.
Firefox:
newsletter-footer-ff
Chrome:
newsletter-footer-chrome

beafialho:
> 26vw looks like this on my screen at 1920x1080. Firefox: newsletter-footer-ff Chrome: newsletter-footer-chrome

Oh, that doesn't look good. Please keep the **18vw** we had before 👍

PR #173, opened 2024-09-02
Add the 70px font size as a custom CSS property in theme.json

### LINKED ISSUE #170: Add additional font sizes as a custom CSS property in theme.json
The 70px font size is used in some patterns, but not used in the typography preset.
Adding more font sizes to the preset would change the font size option to a dropdown, which should be avoided.

Instead of hardcoding the 70px font size in the patterns where it is needed, the theme should use a CSS variable.
That way, if the size value needs to be changed, it can be changed from one place, theme.json.

In theme.json, within the `settings` section, add a new `custom` object and add the font size value.
I suggest using the slug "font-size--huge" and the value 70px. If we want to make this fluid, though, this would be a good time.

Then, replace any instance of the 70px font size in the patterns with the new CSS custom property.

Resources:
https://developer.wordpress.org/themes/global-settings-and-styles/settings/custom/

kishanjasani:
**Description**
Adds custom font size as a custom CSS property. Closes: #170 

**Testing Instructions**

1. After Activate theme, Go to any post or page 
2. RIght click on page and open Inspect element.
3. In that go to `style` section and You will get the custom variable named: `--wp--custom--font-size--huge`  https://prnt.sc/-wgvVIx4wqbn

### DIFF
--- a/patterns/contact-centered-social-link.php
+++ b/patterns/contact-centered-social-link.php
@@ -13,8 +13,8 @@
 ?>
 <!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|80","bottom":"var:preset|spacing|80"},"blockGap":"var:preset|spacing|50"}},"layout":{"type":"constrained"}} -->
 <div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--80);padding-bottom:var(--wp--preset--spacing--80)">
-	<!-- wp:paragraph {"align":"center","fontSize":"xx-large"} -->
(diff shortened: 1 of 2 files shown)


carolinan:
Thank you for working on this.
In theme.json > Styles > Elements > H1 there is also a 70px font size that needs replacing.

kishanjasani:
@carolinan Updated the H1 tag font-size. Thanks!

juanfra (commented):
Thank you for working on this!

I see that the new size is used, so those changes are workin well 🎉 

Though, when viewing the pattern in the front-end it is not respecting the design as the text breaks in three lines. This is happening because the layout of the pattern is taking the max width of 645px, where the intention of the design is that it takes a full width and wrapping the content in the wide width. 

https://github.com/user-attachments/assets/759ef990-27e6-4a7f-b321-3c47d264c1bf

We could move forward with this as is (with the changes that Carolina requested), in that case we'll need to address the layout issues in a new PR.

carolinan:
Yes the width is unrelated to this issue and PR.

PR #176, opened 2024-09-02
Add: CTA - Event RSVP Pattern

### LINKED ISSUE #69: Block Patterns - CTA - Event RSVP
**Description**

🛠️ The goal of this issue is to track the implementation of the `Call to action - Event RSVP` pattern. 
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1779&t=IqDHO3nwfV13Hcvm-1)

CTA_ Event RSVP

beafialho:
**Description**

Adds Block Patterns - CTA - Event RSVP #69

**Screenshots**

https://github.com/user-attachments/assets/0464cfd0-d1d3-40fe-986e-f2020d4f295d


patterns/event-rsvp.php
```diff
@@ -0,0 +1,58 @@
...
+<!-- wp:spacer {"height":"0px","style":{"layout":{"selfStretch":"fixed","flexSize":"100px"}}} -->
+<div style="height:0px" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer -->
+
+<!-- wp:heading {"fontSize":"xx-large"} -->
+<h2 class="wp-block-heading has-xx-large-font-size"><a href="#">RSVP</a></h2>
+<!-- /wp:heading --></div>
+<!-- /wp:group -->
+
+<!-- wp:cover {"url":"http://tt5.local/wp-content/uploads/2024/09/rsvp_image-1.webp","id":2697,"dimRatio":0,"overlayColor":"base","isUserOverlayColor":true,"isDark":false,"layout":{"type":"constrained"}} -->
```
beafialho: Forgot the alt text:

"Close up photo of white flowers on a grey background"

juanfra:
I've adjusted the code for this pattern.

- Added the image to the theme (also the license to the `readme.txt`)
- Adjusted code to use the image.
- Added alt to the image.
- Used the `is-style-section-1` instead of the background color for the bottom-left section.

patterns/event-rsvp.php
```diff
@@ -0,0 +1,75 @@
...
+<!-- wp:group {"align":"full","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull">
+	<!-- wp:group {"align":"full","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"grid","minimumColumnWidth":"70rem"}} -->
+	<div class="wp-block-group alignfull">
+		<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"var:preset|spacing|80","left":"var:preset|spacing|40","right":"var:preset|spacing|40","top":"var:preset|spacing|40"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between","verticalAlignment":"top"}} -->
+		<div class="wp-block-group alignwide" style="padding-top:var(--wp--preset--spacing--40);padding-right:var(--wp--preset--spacing--40);padding-bottom:var(--wp--preset--spacing--80);padding-left:var(--wp--preset--spacing--40)">
+			<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|20"}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"stretch"}} -->
+			<div class="wp-block-group">
+				<!-- wp:heading {"level":1} -->
+				<h1 class="wp-block-heading">Stories, historias, iсторії, iστορίες</h1>
```
carolinan, line 24: Is this a landing page meant for a "page with no title"? Otherwise we need to avoid using the H1's inside the patterns. Because there will be two H1's and the pre-made designs should help the users avoid that.

patterns/event-rsvp.php
```diff
@@ -0,0 +1,75 @@
...
+		</div>
+		<!-- /wp:group -->
+
+		<!-- wp:cover {"url":"<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/image-from-rawpixel-id-8812207.webp","dimRatio":0,"overlayColor":"base","isUserOverlayColor":true,"isDark":false,"layout":{"type":"constrained"}} -->
+		<div class="wp-block-cover is-light">
+			<span aria-hidden="true" class="wp-block-cover__background has-base-background-color has-background-dim-0 has-background-dim"></span><img class="wp-block-cover__image-background" alt="Close up photo of white flowers on a grey background" src="<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/image-from-rawpixel-id-8812207.webp" data-object-fit="cover"/>
+			<div class="wp-block-cover__inner-container">
+				<!-- wp:paragraph {"align":"center","fontSize":"large"} -->
+				<p class="has-text-align-center has-large-font-size"></p>
+				<!-- /wp:paragraph -->
```
carolinan, line 69: I don't mind, but want to double check that this is intended to use a cover block and not another image format.

patterns/event-rsvp.php
beafialho: It would be most suitable for a page with no title, yes. If that's the case, we can make it H2 but using the same (or the most similar) font size.

patterns/event-rsvp.php
beafialho: Yes, this was intentional 👍

carolinan:
I think we should include 'page' in the file name like the landing page patterns in Twenty Twenty-Four.
For the pattern to show up in the modal when a user creates a new template or page, it needs to have this code in the file header:

` * Post Types: page, wp_template`

patterns/event-rsvp.php
```diff
@@ -0,0 +1,75 @@
+<?php
+/**
+ * Title: Event RSVP
+ * Slug: twentytwentyfive/event-rsvp
+ * Keywords: call-to-action, rsvp, event
+ * Categories: call-to-action
```
carolinan, line 6:
```suggestion
 * Categories: call-to-action
 * Post Types: page, wp_template
```

beafialho: Yes, I agree

carolinan:
After testing, I found that the pattern can't be inserted into regular post or page content anymore.
I found that in TT4, there were two almost identical patterns: one for the landing page, and one for the content.
They have different background colors **and different heading levels.**

Should we do the same in TT5?

Issue #147, opened 2024-08-28
Default header: make adjustments to groups and alignments


beafialho:
The current default header has a few details that need to be addressed. 
- Nav items are a group and they extend vertically which does not look good (if we set them to be a row they wrap the intended way)
- The row inside the group wasn't aligned with the other wide templates
- I also adjusted the spacing of the row that contains the navigation block. When there are plugins installed or, for instance social icons, they look better on mobile

Here's the code with my suggested modifications:

```
<!-- wp:group {"style":{"spacing":{"padding":{"right":"0","left":"0"}}},"layout":{"type":"default"}} -->
<div class="wp-block-group" style="padding-right:0;padding-left:0"><!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30","left":"var:preset|spacing|50","right":"var:preset|spacing|50"}}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"space-between"}} -->
<div class="wp-block-group alignwide" style="padding-top:var(--wp--preset--spacing--30);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--30);padding-left:var(--wp--preset--spacing--50)"><!-- wp:site-title {"level":0} /-->

<!-- wp:group {"style":{"spacing":{"padding":{"right":"0","left":"0"},"blockGap":"var:preset|spacing|10"}},"layout":{"type":"flex","flexWrap":"nowrap","justifyContent":"right"}} -->
<div class="wp-block-group" style="padding-right:0;padding-left:0"><!-- wp:navigation {"ref":5,"metadata":{"ignoredHookedBlocks":["woocommerce/customer-account","woocommerce/mini-cart"]},"layout":{"type":"flex","justifyContent":"right","flexWrap":"wrap"}} /-->

<!-- wp:woocommerce/customer-account {"displayStyle":"icon_only","iconStyle":"line","style":{"spacing":{"margin":{"left":"0.5em"}}}} /-->

<!-- wp:woocommerce/mini-cart /--></div>
<!-- /wp:group --></div>
<!-- /wp:group --></div>
<!-- /wp:group -->
```

And a video demoing what I'm mentioning:

https://github.com/user-attachments/assets/94332b6f-2582-499c-acf3-923ee5076e05

carolinan:
[...]

PR #179, opened 2024-09-02
Default header: make adjustments to groups and alignments 

carolinan:
**Description**

This PR adds the updated markup from #147.
With the exception of removing one unused "alignwide". Group blocks who's parent has the default layout type do not have the option to be wide.

**Testing Instructions**

Apply the PR.
Confirm that menus with only a few items do not wrap to multiple lines.
**Bonus:** Test with WooCommerce active. Confirm that the account and cart are on the same line as the menu items.

### DIFF
--- a/patterns/header.php
+++ b/patterns/header.php
@@ -12,14 +12,18 @@
  */
 
 ?>
-<!-- wp:group {"layout":{"type":"constrained"}} -->
-<div class="wp-block-group">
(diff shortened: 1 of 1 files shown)


juanfra (approved):
Thanks, Carolina! 

Code looks good and it works well to me.

https://github.com/user-attachments/assets/f0a701ad-5145-4a68-a583-5c5f31e2eae9

carolinan:
I wonder if the site title should be vertically aligned to the top? Does it look bad when the menu is long enough to be two lines, and they are not aligned?

juanfra:
Personally, I like it as it is now, center-aligned. I'd say that should be a question for Bea, though.

PR #181, opened 2024-09-02
Fix: "Contact centered and social" pattern width.

juanfra:
**Description**

Add full width to parent group and add wrapper to put the content in wide width.

Follow-up of https://github.com/WordPress/twentytwentyfive/pull/173#pullrequestreview-2275041636

**Screenshots**

https://github.com/user-attachments/assets/8a895961-768a-48fb-95c5-08df4e65888e

**Testing Instructions**

1. Create a page.
2. Add the pattern, add a background color to the parent group.
3. View the page in a big screen, confirm that the background is taking the full width and that the content is limited to wide width.

### DIFF
--- a/patterns/contact-centered-social-link.php
+++ b/patterns/contact-centered-social-link.php
@@ -11,19 +11,23 @@
  */
 
 ?>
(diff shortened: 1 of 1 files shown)


beafialho:
Yes, looks good to me!

beafialho (approved):

PR #56, opened 2024-08-19
Add 404 template 

cliffralessio:
**Description**

<-- This pull request introduces a new 404 error page pattern for the Twenty Twenty-Five WordPress theme. -->

**Image License and Source**

<-- Image License: Pexels License-->
<-- Image Source: https://www.pexels.com/-->

**Pexels License Summary**
<-- The images on Pexels can be used for free. -->
<-- Commercial and non-commercial use is allowed. -->
<-- Attribution is not required, but giving credit to the photographer is appreciated. -->

Co-authored-by: cliffralessio <ralessio@git.wordpress.org>

PR #183, opened 2024-09-03
Add 404 template with pattern

carolinan:
**Description**
This replaces https://github.com/WordPress/twentytwentyfive/pull/56 _ only because I was not able to pull down the changes._

I made the following changes:
- Renamed the image file, only removing the` -jpeg` that was at the end of the name, since the theme will use the webp.
- Added a wrapping group block and made the columns wide, as seen in the recording on the original Pr.
- I changed the heading to H1, which meant I needed to hard code the font size value to match the size of the H2 heading.


cliffralessio:
Ok perfect @carolinan  But will I now result on the list of contributors to this project?

carolinan:
yes!

carolinan:
I copied all the credits from the previous PR.

cliffralessio:
Thank you!

Issue #82, opened 2024-08-19
Block Patterns - Events - Poster-like section


juanfra:
**Description**

🛠️ The goal of this issue is to track the implementation of the `Events - Poster-like section` pattern. 
🧑‍💻 The implementation must follow the [pattern creation guidelines](https://github.com/WordPress/twentytwentyfive/blob/trunk/README.md#pattern-creation-guidelines), and should be assigned to its appropriate category.
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1801&t=IqDHO3nwfV13Hcvm-1)

![Events_ Poster-like section](https://github.com/user-attachments/assets/51cf9ed9-b98c-485e-b5ae-70f163932be7)

PR #199, opened 2024-09-03
Add: Poster-like section pattern

beafialho:
**Description**

Adds poster-like section pattern #82.

I figured this one worked nicely if the content inside it had a wide width, instead of full, see how harmonious it looks when using another pattern below.

**Screenshots**

https://github.com/user-attachments/assets/b89bc7a4-58b2-413f-8897-9ef31e66d155


juanfra (approved):
Thanks Bea, it works really well. I agree that it's super nice with the content set to site wide, it works well with the rest.

I've updated the code, added the image and the license for the image. Please give it a try and if everything is ok on your end we should be good to merge.

https://github.com/user-attachments/assets/e2aef2f7-d4f8-4cef-b1fb-0fe31b991b15

beafialho:
@juanfra I can't find the pattern in the patterns sidebar, do you have an idea why that may be happening?

juanfra:
@beafialho it is working for me. Can you check that you're in the right branch? `poster-like-section` is the one you must select, and you should pull the latest changes.

beafialho:
It's working now... Perhaps this is because I was editing a page inside a template. Anyway, this one looks good to merge 👍

juanfra:
That's great; merging this one, then. Thanks!

PR #204, opened 2024-09-03
Add Link in bio with profile, links and wide margins pattern

### LINKED ISSUE #190: Block Patterns - Landing page - Link in bio: Image, heading, short paragraph with wide margins
**Description**

🛠️ The goal of this issue is to track the implementation of the Link in Bio - Heading, paragraph and full-height image on the right pattern.
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1791&t=iMHEZ4xZGwCzs1U9-4)

Captura de ecrã 2024-09-03, às 11 59 56

beafialho:
**Description**

Adds link in bio with profile, links and wide margins pattern #190

**Screenshots**

https://github.com/user-attachments/assets/614aab78-5f1e-4573-9504-47836c961359


patterns/link-in-bio-wide-margins.php
```diff
@@ -0,0 +1,52 @@
...
+ * Categories: banner, about, featured
+ * Description: A link in bio landing page with social links, a profile photo and a brief description.
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+<!-- wp:cover {"overlayColor":"secondary","isUserOverlayColor":true,"isDark":false,"align":"full","style":{"spacing":{"padding":{"right":"var:preset|spacing|80","left":"var:preset|spacing|80","top":"var:preset|spacing|80","bottom":"var:preset|spacing|80"}}},"layout":{"type":"constrained"}} -->
```
juanfra: Is there a specific reason to use a cover for this pattern? I'm wondering if we could use the group block instead of the cover, so that the site admin can use the section styles (this pattern would use the `Style 5`).

juanfra (commented):
Thanks Bea! The pattern is looking good. I have one question, related to the use of cover vs using the group block here.

patterns/link-in-bio-wide-margins.php
beafialho: Yes, great question @juanfra! It's because the cover is meant to be full-height, which I think I forgot to toggle on this pattern. But now that you mention it, it may be possible with group as well.

beafialho: Can you take a look now? I just updated it to use a group block with 100vh.

juanfra: It looks perfect! I believe the horizontal block spacing for the columns needs a bit of adjustment (to look like the original). 

I'm pushing all the changes for you to try it out ✨

beafialho: Thanks, this looks perfect with the changes!

Issue #195, opened 2024-09-03
"Pill" text style


juanfra:
**Description**

![Captura de ecrã 2024-09-02, às 16 01 19](https://github.com/user-attachments/assets/94fb2ffc-9dbc-42f0-9ca2-a1d3d06259b1).

We use a "pill" style for some text elements in several patterns. 

- https://github.com/WordPress/twentytwentyfive/issues/76
- https://github.com/WordPress/twentytwentyfive/issues/77
- https://github.com/WordPress/twentytwentyfive/issues/57
- https://github.com/WordPress/twentytwentyfive/issues/63
- https://github.com/WordPress/twentytwentyfive/issues/64
- https://github.com/WordPress/twentytwentyfive/issues/175

And in the following template as well:

- https://github.com/WordPress/twentytwentyfive/issues/41

This goal of this issue is to track the implementation of the "pill" text style and its usage in different locations.

There's a separate discussion going on about what HTML element would be the best one to use for these "pills" depending on the context, [here](https://github.com/WordPress/twentytwentyfive/issues/150). For the time being, headings and paragraphs looks the most appropriate ones to cover all cases.

There's also a contribution that includes a version of how this text style would look like, it can be found [here](https://github.com/WordPress/twentytwentyfive/pull/118/files#diff-f3d8635069ab04f1108d0fc2002438a75932d712fb5d029d936c5d6423967e93). It's likely that for organization purposes we'll cherry pick that into a different PR so that we can expedite things for the other patterns.

carolinan:
The style is also used for post terms -tags.
The style is different in different style variations. For example the style variation with the grey background has a 4px border radius.

carolinan:
When the block style from #118 is applied to a paragraph or heading, the border is used on the entire length of the block, unless placed in a flex or grid item:

![image](https://github.com/user-attachments/assets/63eb8119-1d6c-4713-bdda-f691378eaca3)

Where a paragraph was intended to be used, I think it may be easier for users to make it look "correct" if a button block is used.
Perhaps this button could be a variation simply called "small button" ?

jasmussen:
The pill-shaped paragraph feels like such a clear design element, and an example of how you can use the block editor to visually express various styles. It seems valid enough to include in patterns, where this is used most often. Is there any reason we can't just include the row wrapper for where this element is needed? The fact that you can unwrap the pill and "break" its appearance is the case for most patterns, including the overlapping gallery patterns and others.

[...]

PR #210, opened 2024-09-04
Create pill style variation

carolinan:
**Description**
This adds a style variation for paragraphs and headings.  **It does not apply the style to any patterns or templates.**
Partial for https://github.com/WordPress/twentytwentyfive/issues/195

**Screenshots**
image

**Testing Instructions**
Add a heading block and apply the "pill" style from the block settings sidebar.
Add a paragraph and apply the "pill" style from the block settings sidebar.
Confirm if the style [matches the design in Figma](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1485&m=dev)

### DIFF
--- a/styles/blocks/pill.json
+++ b/styles/blocks/pill.json
@@ -18,7 +18,7 @@
 		},
 		"spacing": {
 			"padding": {
-				"top": "5px",
+				"top": "0.31rem",
 				"right": "var:preset|spacing|20",
 				"bottom": "0.31rem",
 				"left": "var:preset|spacing|20"
--- a/styles/blocks/pill.json
+++ b/styles/blocks/pill.json
@@ -5,11 +5,11 @@
 	"slug": "pill",
 	"blockTypes": ["core/heading", "core/paragraph"],
 	"styles": {
-        "typography": {
-            "fontSize": "var:preset|font-size|small",
-            "lineHeight": "1",
-            "letterSpacing": "normal"
-        },
+		"typography": {
+			"fontSize": "var:preset|font-size|small",
+			"lineHeight": "1",
+			"letterSpacing": "normal"
+		},
 		"border": {
 			"color": "var:preset|color|contrast",
 			"style": "solid",
--- a/styles/blocks/pill.json
+++ b/styles/blocks/pill.json
@@ -1,0 +1,23 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Pill",
+	"slug": "pill",
+	"blockTypes": ["core/heading", "core/paragraph"],
+	"styles": {
+		"border": {
+			"color": "var:preset|color|contrast",
+			"style": "solid",
+			"width": "1px",
+			"radius": "20px"
+		},
+		"spacing": {
+			"padding": {
+				"top": "5px",
+				"right": "var:preset|spacing|tiny",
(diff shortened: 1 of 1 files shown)


beafialho:
Thank you @carolinan! This is what I'm currently seeing:

Captura de ecrã 2024-09-04, às 09 57 54

Can we make the border wrap the heading text, so there isn't the extra spacing that extends to the width's edges? Here's what I mean:

Captura de ecrã 2024-09-04, às 10 00 16

carolinan:
yeah that is the $100 question, is anyone able to add that to the style?

styles/blocks/pill.json
```diff
@@ -0,0 +1,28 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Pill",
+	"slug": "pill",
+	"blockTypes": ["core/heading", "core/paragraph"],
+	"styles": {
+        "typography": {
```
juanfra: Let's indent this one using tabs.

styles/blocks/pill.json
```diff
@@ -0,0 +1,28 @@
...
+        },
+		"border": {
+			"color": "var:preset|color|contrast",
+			"style": "solid",
+			"width": "1px",
+			"radius": "20px"
+		},
+		"spacing": {
+			"padding": {
+				"top": "5px",
```
juanfra:
```suggestion
				"top": "0.31rem",
```

styles/blocks/pill.json
```diff
@@ -0,0 +1,28 @@
...
+				"right": "var:preset|spacing|20",
+				"bottom": "5px",
```
juanfra:
```suggestion
				"bottom": "0.31rem",
```

styles/blocks/pill.json
```diff
@@ -0,0 +1,28 @@
...
+        "typography": {
+            "fontSize": "var:preset|font-size|small",
+            "lineHeight": "1",
+            "letterSpacing": "normal"
+        },
+		"border": {
+			"color": "var:preset|color|contrast",
+			"style": "solid",
+			"width": "1px",
+			"radius": "20px"
```
juanfra:
```suggestion
			"radius": "999px"
```

juanfra (commented):
Thanks, Carolina!

https://github.com/user-attachments/assets/d01d3cde-0510-4ea3-9180-c4aa36f1e601

I believe we should use `rem` units here, and add a "safe" border radius so that the style adapts well to other font-sizes, which can happen. So that the spacing and the radius are still working.

Regarding the text-wrapping, I believe that for the time being we don't have a better and safer option than using this style inside a "stack".

carolinan:
> I believe we should use rem units here, and add a "safe" border radius so that the style adapts well to other font-sizes, which can happen. So that the spacing and the radius are still working

Yes good idea.

beafialho:
> Regarding the text-wrapping, I believe that for the time being we don't have a better and safest option that using this style inside a "stack".

Good idea. Will the style contain that stack though, or do we need to add it to the heading/paragraph when we're building the pattern?

carolinan:
In the patterns.

juanfra (approved):
Thanks, Carolina! It looks good to me. I'll merge this one so that we can start using it in the different patterns.

Issue #133, opened 2024-08-27
Add alternative templates for News Blog


beafialho:
[Figma File](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3155-1265&t=bCFceo6KG1JJBgYQ-4)

These templates are all alternative templates to the default set but number one should definitely be built. The others we can, depending on the time we have, reconsider building or not.

**Alternative One**: #42

**Alternative Two**

- [x] Blog

![Captura de ecrã 2024-09-04, às 11 04 36](https://github.com/user-attachments/assets/86180974-5cd1-4fd4-a7f8-7cf92fef2566)

**Alternative Three**

- [x] Blog
- [x] Single

![Captura de ecrã 2024-08-27, às 16 21 26](https://github.com/user-attachments/assets/331fb930-441b-4055-8c4a-c45f53af7367)

**Alternative Four**

- [x] Single

![Captura de ecrã 2024-08-27, às 16 21 32](https://github.com/user-attachments/assets/2f4851aa-ca3c-4b1b-b986-79c0ea31ce02)

beafialho:
@carolinan could you also assign this one to me please?

PR #215, opened 2024-09-04
Add alternative templates/patterns for News Blog

beafialho:
**Description**

Adds alternative templates/patterns for News Blog mentioned in #133

There may be some discrepancies from the original Figma, this was intentional and built as looking good in the editor and front end.


juanfra:
I've been playing a bit with this one:

- Moved the templates to the correct locations.
- Fixed lint issues.
- Added docblocks.
- Update the single post ones to use the comments pattern.

@beafialho @carolinan can you please take a look at the changes and confirm that things are in place? Thanks

patterns/news-blog-with-sidebar-template.php
```diff
@@ -0,0 +1,115 @@
...
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+<!-- wp:template-part {"slug":"header","tagName":"header","area":"header"} /-->
+
+<!-- wp:group {"className":"is-style-default","style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group is-style-default" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50)"><!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"left":"var:preset|spacing|50"}}}} -->
+<div class="wp-block-columns alignwide"><!-- wp:column {"width":"75%"} -->
+<div class="wp-block-column" style="flex-basis:75%"><!-- wp:query {"queryId":41,"query":{"perPage":1,"pages":0,"offset":0,"postType":"post","order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false},"metadata":{"categories":["posts"],"patternName":"core/query-standard-posts","name":"Standard"}} -->
```
carolinan, line 20: Does the metadata need to be here for this core pattern to be used?

carolinan:
"News blog with featured posts grid" Is missing an H1 and the categories are above the headings, so screen reader users that navigate via headings will not get the information about the category. Instead the category is read as belonging under the previous heading, and this would be confusing since the actual category might be something else.

The same posts are displayed twice. First in the main query and then in the list under the "Architecture" heading.

----

"News blog with sidebar" is missing an H1: All headings have the same level which makes it difficult to navigate and prioritize the content. The heading "Latest" in the sidebar has the same heading level; the same importance, as all the blog posts.

The main post is displayed three times, the other posts are displayed twice.
The list below the main post, and the sidebar, has the same posts. I don't think there is a benefit to repeating these links?

In the list below the main post, the posts titles are not linked.

----
"Offset post without featured image" is missing an H1.
I _believe_ the border above the next and previous posts links is meant to be wide? It is full width.


----
"Post with left aligned content"
I _believe_ the border above the next and previous posts links is meant to be wide? It is full width.

beafialho:
> "News blog with featured posts grid" Is missing an H1 and the categories are above the headings, so screen reader users that navigate via headings will not get the information about the category. Instead the category is read as belonging under the previous heading, and this would be confusing since the actual category might be something else.

I have a doubt regarding this, can't the site title be H1?

Captura de ecrã 2024-09-10, às 10 56 35

beafialho:
Thank you for reviewing, I submitted some changes according to your feedback.

carolinan:
> I have a doubt regarding this, can't the site title be H1?

Here are some resources I trust about heading levels:
https://make.wordpress.org/accessibility/handbook/content/using-headings-in-content/#what-are-headings
https://webaim.org/techniques/headings/
https://www.a11yproject.com/posts/how-to-accessible-heading-structure/

The site title in the header is not a good H1 for a few reasons:
It is identical on every page, so it is not for the unique content on the current page. It does not describe the content on the current page.

From a block theme perspective, a single pattern or template part can not change the HTML element for the title _conditionally_: it can't use an H1 on one specific page, and a paragraph on another. This was possible in classic themes.
Block themes can use **multiple** patterns or parts with different HTML elements, and use those as needed. But using the site title as the H1 is still not ideal for the other reasons listed.

carolinan:
### News blog with featured posts grid
 - [x] 1 `H1`, no repeated posts.

I did not test how the page would look if there were less than five posts.
I think it may be confusing that the "no results" message would show four times?

### News blog with sidebar
 - [x] 1 `H1`, no repeated posts. The sidebar content uses H3.
 
 ### Offset post without featured image
 - [x] 1 `H1`, border width.
 
### Post with left aligned content
The border above the next and previous posts link is full width, but perhaps that is the design choice.

----

### Changes to existing templates

#### News blog home
- [x]  Categories were moved below the headings.

beafialho:
> I did not test how the page would look if there were less than five posts.
I think it may be confusing that the "no results" message would show four times?

I think this is expected since this is a template to be used when there are many posts.

> The border above the next and previous posts link is full width, but perhaps that is the design choice.

That is correct, the border here is intended to be full width.

Thank you for reviewing!

PR #211, opened 2024-09-04
Add Event schedule pattern

beafialho:
**Description**

Adds Event schedule pattern #85

**Screenshots**

https://github.com/user-attachments/assets/672e166c-c804-4558-8587-1ac5df29d245

PR #225, opened 2024-09-05
Add: Instagram grid pattern

### LINKED ISSUE #222: Add pattern: Instagram grid
🛠️ The goal of this issue is to track the implementation of the Instagram grid pattern.

[Figma](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3177-1799&m=dev)

image

beafialho:
**Description**

Adds instagram grid pattern #222

**Screenshots**

Captura de ecrã 2024-09-05, às 11 39 53


juanfra (commented):
Thanks Bea! I've adjusted the code, added images and their respective licenses.

The one thing I noticed is that the first item does not look squared on mobile (tested in Chrome and Firefox).

https://github.com/user-attachments/assets/a2c617ba-b81c-4578-8ecc-e0943d69c183

beafialho:
Oh! That one is not supposed to look like that. I did not want to use a cover because I wanted to select the section style color. Can we replace the _group_ block containing the "Instagram @example" text with a _cover_ block? That way, it behaves as intended:

Captura de ecrã 2024-09-06, às 15 18 39

carolinan:
3 of these images are also in https://github.com/WordPress/twentytwentyfive/pull/211
Can we re-use them and use more neutral file names?

beafialho:
I think that would be ok 👍 The pattern has a square aspect ratio so it should not alter anything.

juanfra:
Sounds good - I've updated the pattern to use the cover block with a background color, and I also changed the filenames so that these images can be used in the other PR.

carolinan:
> I also changed the filenames so that these images can be used in the other PR.

I forgot and already merged the other PR 😞

carolinan:
The file named hibiscus-flower-square.webp is not the hibiscus but the meadow with the flowers and the tree in the background,
the existing hibiscus flower image, grid-flower-1.webp, was not replaced.

juanfra:
Thanks, @carolinan - Good catch about the image mismatch in the readme. I've updated that.

I've updated the `agenda-img-4.webp` image, because it wasn't aligned to the other images in the other pattern (it was too small).

PR #220, opened 2024-09-05
Register the page pattern category

ibrahim-kardi:
**Description**
Fixed the opened issue #219

**Screenshots**
image

**Testing Instructions**
When creating a new pattern -> in category field type 'pages'

PR #230, opened 2024-09-05
Wrap functions in `function_exists` conditional for consistency.

juanfra:
**Description**

In https://github.com/WordPress/twentytwentyfive/pull/220/files we've added a new function and we added it with `function_exists` conditional. To be consistent, I'm adding the same conditional to the other functions in the `functions.php` file.

**Screenshots**

N/A

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -11,79 +11,103 @@
 
 /**
  * Enqueue editor-style.css in the editors.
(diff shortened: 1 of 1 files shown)


carolinan (approved):
Works well, thank you.

PR #231, opened 2024-09-05
Update heading sizes and default home template

beafialho:
**Description**

Currently, the default home template looks like this if we add a post with some headings in the content:

Captura de ecrã 2024-09-05, às 16 24 05

It's not a good look that the post title is smaller than the subsequent headings from the post content.

So I tinkered with the heading sizes in global styles to see how we could solve this issue. I also checked with the patterns and although the scale of the headings in some of them is a bit smaller, I think they still look good. This is how it looks with the updated suggestion:

Captura de ecrã 2024-09-05, às 16 28 20

The sizes are the following:

**H1: 2.62rem
H2: 1.8rem
H3: 1.64rem
H4: 1.5rem
H5: 1.25rem
H6: 0.875rem (Bold, uppercase, 1.4px letter spacing)**

I also changed the font size of the "Blog" heading in the actual template.

Once these are reviewed, I'll update these heading sizes on the Figma file for coherency.

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -2,8 +2,8 @@
 
 <!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
 <main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
-<!-- wp:heading {"textAlign":"left","level":1,"className":"is-style-default"} -->
-<h1 class="wp-block-heading has-text-align-left is-style-default">Blog</h1>
+<!-- wp:heading {"textAlign":"left","level":1"} -->
+<h1 class="wp-block-heading has-text-align-left">Blog</h1>
 <!-- /wp:heading -->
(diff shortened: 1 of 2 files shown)


templates/home.html
```diff
@@ -2,9 +2,9 @@
 
 <!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
 <main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
-	<!-- wp:heading {"level":1,"fontSize":"xx-large"} -->
-	<h1 class="wp-block-heading has-xx-large-font-size">Blog</h1>
-	<!-- /wp:heading -->
+<!-- wp:heading {"textAlign":"left","level":1,"className":"is-style-default"} -->
+<h1 class="wp-block-heading has-text-align-left is-style-default">Blog</h1>
```
carolinan:
```suggestion
<!-- wp:heading {"textAlign":"left","level":1"} -->
<h1 class="wp-block-heading has-text-align-left">Blog</h1>
```

juanfra:
I believe we need to make a pass on the different patterns to get the sizes matching.

PR #237, opened 2024-09-05
Move font family definitions into respective typography presets.

### LINKED ISSUE #233: Move the style variation specific font families from the main theme.json to the variation json file.
## What?
Move the style variation specific font families from the main theme.json to the variation JSON file.

## Why?
Currently, the font families must be defined in the main theme.json file to load as expected in the editor when the style variation is switched. If this issue https://github.com/WordPress/gutenberg/issues/59965 is fixed (in progress https://github.com/WordPress/gutenberg/pull/65019), we could avoid this 'hack' that's producing the printing of not needed CSS in the frontend. So, this change would be a tiny performance improvement, and it could lead to simpler-to-define typography variations.

jffng:
## What?
This PR moves the font family definitions to their respective typography preset files. 

Needs https://github.com/WordPress/gutenberg/pull/65019 to be merged to work as expected.

## How?
- Each style variation or typeset file defines the font families used in that variation/typeset.
- Main `theme.json` only includes 2 font families used in the default styles (Manrope and Fira Code).

## Why?
Fixes: https://github.com/WordPress/twentytwentyfive/issues/233


jasmussen:
Hey @jffng, we're nearing beta 1. Is this something that has to happen before that date?

jffng:
> Is this something that has to happen before that date?

I don't think so, since the related Gutenberg issue is considered a bug.

matiasbenedetto:
I updated this branch incorporating the latest changes made on trunk. I moved all font family definitions that are not used in the default styles out of the main `theme.json` file and included them on each variation/typeset file.

matiasbenedetto:
This screencast compares the behavior of
| Left | Right |
| --- | --- |
| WP 6.6 + Gutenberg with https://github.com/WordPress/gutenberg/pull/65019 and TT5 with this PR (https://github.com/WordPress/twentytwentyfive/pull/237) |  WP 6.7 Beta2 |

Both screens should look the same in terms of typography. The video shows that the fonts are loading correctly when defined in a style variations and typesets (https://github.com/WordPress/twentytwentyfive/pull/237) when https://github.com/WordPress/gutenberg/pull/65019 is in place.


[Screencast from 09-10-24 18:22:50.webm](https://github.com/user-attachments/assets/6c16fa59-6a71-4743-bc37-025f0df85b15)

matiasbenedetto:
I merged the latest changes from `trunk` branch and resolved the conflicts.
Now https://github.com/WordPress/wordpress-develop/pull/7581 has been merged into core, and https://github.com/WordPress/gutenberg/issues/59965 has been closed; I'm removing the blocked tag.

I think the PR is in good shape for review.

juanfra (approved):
Thank you. I've tested this saving each style variation / typography present and viewing in the front end, and it's working well. Code also looks good.

https://github.com/user-attachments/assets/b465a0fc-b0ac-444f-afc8-26da6afcc60b

juanfra:
I've just rebased after we got [this PR merged](https://github.com/WordPress/twentytwentyfive/pull/603), where we introduced variable fonts. There were some conflicts in `theme.json` and I updated the font family definitions in the typography presets to use the variable fonts.

@matiasbenedetto @carolinan @beafialho I'd appreciate it if you could look at this and confirm everything is working as expected. It is working as expected on my end. Thanks 🙌

carolinan:
To me this is a design decision, since it will change where the font families are available to be selected.

beafialho:
I'm not sure if my testing is showing the right things, but here's what I noticed:

- When I select Afternoon, Midnight and Sunrise combined variations, I don't see the typography variations selected, like it happens with the others:

https://github.com/user-attachments/assets/ecb8a805-b557-43f3-af49-39e6a2d22287

https://github.com/user-attachments/assets/39b0b39d-2786-4cdb-b2a2-41d5ec533273

- I am also noticing that the Afternoon combined variation and its respective typography set (which should look the same) don't look the same:

https://github.com/user-attachments/assets/d834532e-b094-460e-936d-37c12002d2c9

carolinan:
@beafialho I had not even noticed that when you select the combined variations (theme style variations) the color palette and typography preset gets a visible outline. I did not know that was a thing.

You are describing that when some variations are selected, the matching color palette and typography presets are not selected, and I can reproduce that even without this PR. So there might be a problem that is unrelated to the changes in this PR.

beafialho:
> @beafialho I had not even noticed that when you select the combined variations (theme style variations) the color palette and typography preset gets a visible outline. I did not know that was a thing.

I had not checked with proper eyes either @carolinan, but as I tested them now, I expected that to be the behaviour.

> So there might be a problem that is unrelated to the changes in this PR.

I also suspected that, but I also wasn't sure if it might be just this PR, since it was created way before the changes in #564 which improved the typography variations.

carolinan:
This is what I meant when I tried to describe that I consider this a design decision.
Before, without this PR; the font options on all variations contain these fonts:
Screenshot 2024-10-28 121147
After, the font options in each variation is different.
Afternoon:

image

I am not trying to say that the "before" is better (it might be overwhelming), just highlighting the difference.
Only loading the required font files should be a performance improvement.

matiasbenedetto:
> This is what I meant when I tried to describe that I consider this a design decision.

Yes, I think it's the proper design decision. From my perspective, rendering all the fonts from all the style variations and letting the user select them for page elements doesn't make sense. The variations are meant to be alternative styles that could be replaced but not combined. A design with nine different font families is difficult to harmonize, and I don't think it is the idea we would like to implement.

beafialho:
Thank you for clarifying @carolinan!

As a designer, I have a preference for seeing all the font families displayed, regardless of the variation I'm using, for convenience but also logic of seeing all the fonts shipped with the theme in one list. However, I do see how displaying that list in the sidebar below the presets can be overwhelming and even lead to unwanted, less than ideal page/element specific settings. Since it is possible and easy to add Google Fonts and even upload them, which gives users the option of adding whatever font they want, I think this PR makes sense to be merged.

juanfra:
Thank you all! I'm merging this one.

PR #243, opened 2024-09-06
Add default style for the separator block

### LINKED ISSUE #241: Separator: Improve appearance
Currently, the separator is looking a bit rough. I'm aware this one can't be styled in the editor, but can we style it some other way? Here's how it's looking vs. how it should ideally look:

Captura de ecrã 2024-09-06, às 11 35 31

It should ideally have a `border: 1px` and use the `var:preset|color|opacity-20`.

carolinan:
**Description**
This PR adds a default style for the separator block, using a 1px border with the color opacity-20.
I copied this solution from Twenty Twenty-Four, so I added the credits to the readme.txt file.

Closes https://github.com/WordPress/twentytwentyfive/issues/241

**Screenshots**
image

**Testing Instructions**
Apply the PR and insert the separator block.
Try the different widths and styles of the block.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -30,6 +30,11 @@ but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 GNU General Public License for more details.
 
+This theme incorporates code from:
+
+Twenty Twenty-Four WordPress Theme, (C) 2023 WordPress.org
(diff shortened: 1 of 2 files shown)


beafialho:
Thank you, looks perfect 🚀

juanfra (approved):
Looks good to me! 🚀

https://github.com/user-attachments/assets/04f8356e-b6d2-4f02-98d5-c18deeabbf8f

I'm not familiar with how it should be when it comes with crediting other default themes, so I'll defer to you.

Issue #29, opened 2024-08-15
Use the recommended `var:preset|$feature|$slug` syntax to reference presets in `theme.json`


germanfrelo:
[TT4' theme.json only used the CSS syntax](https://github.com/search?q=repo%3AWordPress%2Ftwentytwentyfour+path%3A**%2Ftheme.json+%22var%28--wp--preset--%22&type=code) to reference presets. It would be great to use the `var:preset|$feature|$slug` syntax in this TT5 theme to encourage its use, as recommended in "[Applying presets as styles – Using Presets – Theme Handbook | Developer.WordPress.org](https://developer.wordpress.org/themes/global-settings-and-styles/styles/using-presets/#applying-presets-as-styles)":

> Technically, you can reference presets using the CSS syntax of `var( --wp--preset--{$feature}--{$slug} )`. But the WordPress `var:preset|$feature|$slug` syntax works much better and always appears correctly throughout the interface in the WordPress admin. Save the CSS syntax for when you are actually writing CSS.

carolinan:
There are still a couple of places in the section style files (section-1.json etc) inside the styles folder that needs to be updated.

germanfrelo:
I think this can be closed since last PR https://github.com/WordPress/twentytwentyfive/pull/248?

PR #248, opened 2024-09-06
Refactor: Replace `var(--wp--preset--feature-slug)` with `var:preset|…

KevinTaron:
…$feature|$slug` syntax

Updated all instances where `var(--wp--preset--feature-slug)` was used to the recommended `var:preset|$feature|$slug` syntax for consistency and improved maintainability.

Solving the issue #29


carolinan (approved):
I test ran the change and the new values work as excepted, thank you.

PR #251, opened 2024-09-07
Remove the class name is-style-default#249

### LINKED ISSUE #249: Remove the class name "is-style-default"
**Description**
The CSS lass name "is-style-default" is added by the block editor if a user selects a block style variation and then switches back to the default style.

Some patterns now have this class name in the markup, and it should be removed because it is not needed.
After updating the markup, the pattern must be tested in the editor so that there are no block validation errors, because typos happen easily. 😃 

**Example**
Image block:
`"className":"wp-block-image size-large is-style-default"` can be changed to `"className":"wp-block-image size-large"`.
`<figure class="wp-block-image size-large is-style-default">` can be changed to `<figure class="wp-block-image size-large">`

```
<!-- wp:image {"aspectRatio":"1","scale":"cover","linkDestination":"none","className":"wp-block-image size-large is-style-default"} -->
<figure class="wp-block-image size-large is-style-default"><img src="<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/location.webp" alt="The business location" /></figure>
<!-- /wp:image -->
```

ibrahim-kardi:
**Description**
This issue resolved: Remove the class name "is-style-default" #249
"is-style-default" is added by the block editor if a user selects a block style variation and then switches back to the default style. I have removed this from all patterns where it was used.

### DIFF
--- a/patterns/banner-with-description-and-images-grid.php
+++ b/patterns/banner-with-description-and-images-grid.php
@@ -11,8 +11,8 @@
  */
 
 ?>
-<!-- wp:group {"align":"full","className":"is-style-default","layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignfull is-style-default">
+<!-- wp:group {"align":"full","layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull">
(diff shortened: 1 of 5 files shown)


juanfra (approved):
Thank you! Looks good to me.

rejaulalomkhan:
Yeah.! That's great. It working for me. Thanks @ibrahim-kardi

Issue #193, opened 2024-09-03
Add the "combined" global style varitations


carolinan:
**Is your feature request related to a problem? Please describe.**

The theme now has typography presets, color palette presets, and section styles.
This issue is for creating the style variations that combine different presets: In other words, the json files for the style variations that will be available next to the default style here at the top of the Styles panel:

![image](https://github.com/user-attachments/assets/97238e1f-5000-4137-9126-09d3f6c6abd1)

[Here are the current Figma designs](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3155-1729)

- [x] Evening - Manrope + Manrope (Dark mode)  _(kishanjasani)_
- [x] Noon - Beirut & Literata 60P _(shail-mehta)_
- [x] Dusk - Volkorn and Fira code     _(carolinan)_
- [x] Afternoon - Platypi + Ysabeau Office _(kishanjasani)_
- [x] Twilight - Roboto Slab + Manrope _(beafialho)_
- [x] Morning - Literata 60pt + Ysabeau Office _(juanfra)_
- [x] Sunrise - Platypi + Literata 60pt _(juanfra)_
- [x] Midnight - Literata 60pt + Fira Sans _(beafialho)_

carolinan:
I'd like to work on Dusk.

beafialho:
I'd like to work on Twilight and Midnight.

juanfra:
I'm not a morning person, but I'll make an exception; I take Morning and Sunrise.

kishanjasani:
I'd like to work on `Afternoon - Platypi + Ysabeau Office` and `Evening - Manrope + Manrope (Dark mode)`.

shail-mehta:
I would like to work on `Noon - Beirut & Literata 60P`  style variation.

carolinan:
Hi @shail-mehta @kishanjasani do you still want to work on these? Or can anyone work on these styles?

beafialho:
Thanks for double checking @carolinan, I was going to ask this. I'm happy to handle these as well 🙋🏻‍♀️

juanfra:
Given the priority of this issue and the recent activity, I'd prefer if Bea could start with this. Also, considering that she'll be out starting Friday.

beafialho:
I'll get started on these given that it is time-sensitive.

juanfra:
Thanks, Bea. Thank you all!

[...]

PR #261, opened 2024-09-09
Add: Morning style variation.

juanfra:
**Description**

Add the Morning (Literata 60pt + Ysabeau Office) global style variation.
Partial for https://github.com/WordPress/twentytwentyfive/issues/193

**Known limitations**

- Many font sizes don't match the presets. They were hardcoded in this PR.
- The post author section has a hardcoded font preset in the template that doesn't match the design. So the author size is different than the text that says "Written by".
- The font size for the comments section looks small in the context, but there's no design for the comments for this variation.

**Screenshots**

Screenshot 2024-09-09 at 18-30-53 Style Variations – Stories®

**Testing Instructions**

1. Go to Appearance > Editor > Styles
2. At the top of the styles panel, select "Morning".
3. Save.
4. Compare the design with the design in Figma


beafialho:
@juanfra I noticed the titles are looking different than the headings, they seem to have specific styling. If you disable them, the appearance looks right:

https://github.com/user-attachments/assets/32529e43-04b9-41ca-aa00-f492fc63a5d3

The font sizes are not looking very good. I think we should reset all the heading sizes in this variation to use the sizes from the default one, as they look much better.

carolinan (changes requested):
### Style 2
Headings use the text color `#191919 `(contrast) against the background `#182949` (accent 3). This has a contrast ratio of
1.21:1. It needs to be 4.5:1.

I believe this is meant to use base, to match the paragraph text color.

juanfra:
Thanks, Carolina. Good catch. I've updated the variation with that change.

styles/morning.json
```diff
@@ -0,0 +1,338 @@
...
+			"link": {
+				"color": {
+					"text": "var:preset|color|contrast"
+				}
+			}
+		},
+		"variations": {
+			"post-terms-1": {
+				"typography": {
+					"fontSize": "16px"
```
carolinan, line 244: Does this need to be hardcoded? Is there no matching responsive preset size?

carolinan:
Feedback: I am struggling to read font size preset `small` which is 14px, up to preset `medium` 18px with this font over the grey background.
The text color and background color has a contrast ratio of 4.66:1 which is within the WCAG AA level requirement, but for me, a higher contrast or even better a larger size would help.

This is even more appearant with the new zoom out.

Both the default and Dusk are much more readable to me because of the different size of these fonts.

styles/morning.json
juanfra: There's no preset for 16px, which is the size of the font size for this particular element in the variation. We have `--wp--preset--font-size--normal` from Gutenberg, which is 16px, but we can't use that.

If we don't want to hardcode the value, we could use medium (18px) or small (14px) alternatives.

carolinan:
https://github.com/user-attachments/assets/4e86f740-679d-4011-a521-fe750ff3be7c

carolinan:
You can see just how much smaller this font is by the number of menu items that fit on the same row.

juanfra:
Thanks for the feedback, Carolina.

The issues you highlighted are related to the size of the Ysabeau font, compared to fonts used in other variations. The menu item size will happen the other way around with variations like Twilight or Midnight, where the font size for those elements is bigger.

I believe this is a conversation we want to have with design, as some of these elements are probably intended to be like that. Given the time constraints of beta 1, if these are the only blockers, we could merge this and have a new issue to discuss what you highlighted and get @beafialho and @jasmussen in the conversation.

styles/morning.json
```diff
@@ -0,0 +1,338 @@
...
+				},
+				"typography": {
+					"fontSize": "1.5rem",
+					"fontWeight": "500",
+					"letterSpacing": "-0.18px"
+				},
+				"elements": {
+					"cite": {
+						"typography": {
+							"fontSize": "1.12rem",
```
carolinan: I believe this needs to use em if it is supposed to be relative to the size of the quote.

styles/morning.json
carolinan: I'd always go for the larger :)

jasmussen:
Zoom out, give or take any changes that comes to this, should not dictate any font size rules in the theme. 

The core of the issue, though: is the gray on light gray font legible at 14px? For me it is, so I'd be tempted to echo Juan and merge this for Beta 1 and then revisit the font sizes (or perhaps the font _weights_) after the beta period. Notably, Bea is AFK this week, a vacation planned months before the 6.7 cycle even had dates attached, and I'd appreciate her quick gut-check when she gets back.

How does that sound?

carolinan:
Nobody is pressuring anyone to not have a vacation ❤️

jasmussen:
> Nobody is pressuring anyone to not have a vacation ❤️

Oh, I didn't mean to imply otherwise, and I supremely agree!

styles/morning.json
juanfra: The intention was to have it in `rem` to match the design (~18px). Now that I see this, we can use the medium preset. Unless you prefer to have the `em` value.

Screenshot 2024-09-24 at 11 24 56

styles/morning.json
juanfra: Should we run this by Bea once she's back?

juanfra:
> How does that sound?

That sounds great to me. 

I'd prefer not to iterate on the designs now and have Bea (after she's back from her well-deserved time off) participate in the discussion, as there might be some reasoning behind some of the selections (sizes, colors, etc.). We should keep track of all these concerns so we can return to them.

carolinan:
Please remember that the beta 1 date is not something we have the power to move.

PR #124, opened 2024-08-26
Podcast Hero Banner

iamsam2e:
Apologies. I ended up ruining the PR #105 with the wrong rebase, I am in the process of setting up a new desktop and got things mucked up.

I have now done a new PR with the changes and feedback from the first PR, apologies for this miss... error on my part

----
Closes https://github.com/WordPress/twentytwentyfive/issues/91
Figma link: 
https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3181-2243&node-type=SYMBOL&t=KDN0XAoLHxb8rx5d-0

PR #263, opened 2024-09-09
Add: "Hero podcast" pattern.

### LINKED ISSUE #91: Block Patterns - Hero - Podcast
**Description**

🛠️ The goal of this issue is to track the implementation of the `Hero - Podcast` pattern. 
🎨 [Figma design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3181-2243&t=IqDHO3nwfV13Hcvm-1)

Hero_ Podcast

juanfra:
**Description**

- Taking the pattern from https://github.com/WordPress/twentytwentyfive/pull/124 and adding changes, as I can't take over the fork.
- ~Setting the title from an H2 to an H1, although we need to review the sizes after https://github.com/WordPress/twentytwentyfive/pull/231 to make the patterns accessible, but respecting the designs at the same time.~

⚠️Note: We need to review the heading sizes after https://github.com/WordPress/twentytwentyfive/pull/231, the H2 looks too small compared to the design. And the size that matches is what now lives in the H1. But using an H1 would compromise the a11y of the pattern.

**Screenshots**

https://github.com/user-attachments/assets/3d3f2154-74d1-4d8e-97da-2f8341869d48

**Testing Instructions**

1. Create a page.
2. Add the pattern.
3. Test that the pattern looks like the design.


carolinan:
In what way does using the H1 make the pattern accessible, is this a full page landing page? I don't think it is listed as one in Figma?
Using this pattern in content would mean that there would be two H1's.

juanfra:
> In what way does using the H1 make the pattern accessible, is this a full page landing page? I don't think it is listed as one in Figma? Using this pattern in content would mean that there would be two H1's.

Thanks for the comment @carolinan. That's precisely what I mentioned in the description. Since we have https://github.com/WordPress/twentytwentyfive/pull/231, all the sizes look different. 

So, now all the H2s look too small compared to all the pattern designs, and the correct size lives in the implementation of the H1. That's what I added to the description: `we need to review the sizes after ... to make the patterns accessible, but respecting the designs at the same time.`

carolinan:
I don't understand. Having two H1's does not make it more accessible. 
The level can't be changed to H1 only because the font size of the H2 does not match.

juanfra:
@carolinan I'm setting the H1 only to respect the design, and highlighting that we need to review the sizes so that we don't have a11y issues, because of the reason that you have just mentioned 🙂

I will set it to H2 again, but we need to have this conversation somewhere. Because now all the patterns look radically different from Figma.

beafialho:
We should make sure we're using the correct headings in each pattern which, in the case of this one, is H2. I've set all the new sizes in Figma, so they shouldn't look _radically_ different. **What I suggest in cases like this one, where H2 looks rather small, is to keep the H2, but set it to font size preset XXL.** In some cases, size XL might look better. Let's approach case by case.

I suggest we replace the buttons for pill style paragraphs to better match the design.

Here's my suggestion:

```
<!-- wp:group {"metadata":{"categories":["featured","hero"],"patternName":"twentytwentyfive/hero-podcast","name":"Hero podcast"},"align":"full","className":"is-style-section-1","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60","left":"var:preset|spacing|60","right":"var:preset|spacing|60"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
<div class="wp-block-group alignfull is-style-section-1" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--60);padding-right:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60);padding-left:var(--wp--preset--spacing--60)"><!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|60","left":"var:preset|spacing|60"}}}} -->
<div class="wp-block-columns alignwide"><!-- wp:column {"width":"40%"} -->
<div class="wp-block-column" style="flex-basis:40%"><!-- wp:image {"sizeSlug":"large","linkDestination":"none"} -->
<figure class="wp-block-image size-large"><img src="http://tt5.local/wp-content/themes/twentytwentyfive/assets/images/hero-podcast.webp" alt="Picture of a person"/></figure>
<!-- /wp:image --></div>
<!-- /wp:column -->

<!-- wp:column {"verticalAlignment":"center","width":"60%"} -->
<div class="wp-block-column is-vertically-aligned-center" style="flex-basis:60%"><!-- wp:heading {"fontSize":"xx-large"} -->
<h2 class="wp-block-heading has-xx-large-font-size">The Stories Podcast</h2>
<!-- /wp:heading -->

<!-- wp:paragraph {"style":{"spacing":{"padding":{"top":"var:preset|spacing|30"}}}} -->
<p style="padding-top:var(--wp--preset--spacing--30)">Storytelling, expert analysis, and vivid descriptions. The Stories Podcast brings history to life, making it accessible and engaging for a global audience.</p>
<!-- /wp:paragraph -->

<!-- wp:heading {"level":3,"style":{"spacing":{"padding":{"top":"var:preset|spacing|30"}}},"fontSize":"large"} -->
<h3 class="wp-block-heading has-large-font-size" style="padding-top:var(--wp--preset--spacing--30)"><strong>Subscribe on your favorite platform</strong></h3>
<!-- /wp:heading -->

<!-- wp:group {"layout":{"type":"flex","flexWrap":"nowrap"}} -->
<div class="wp-block-group"><!-- wp:paragraph {"className":"is-style-pill"} -->
<p class="is-style-pill"><a href="#">Youtube</a></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"className":"is-style-pill"} -->
<p class="is-style-pill"><a href="#">Apple Podcasts</a></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"className":"is-style-pill"} -->
<p class="is-style-pill"><a href="#">Spotify</a></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"className":"is-style-pill"} -->
<p class="is-style-pill"><a href="#">Pocket Casts</a></p>
<!-- /wp:paragraph -->

<!-- wp:paragraph {"className":"is-style-pill"} -->
<p class="is-style-pill"><a href="#">RSS</a></p>
<!-- /wp:paragraph --></div>
<!-- /wp:group --></div>
<!-- /wp:column --></div>
<!-- /wp:columns --></div>
<!-- /wp:group -->
```

juanfra:
> What I suggest in cases like this one, where H2 looks rather small, is to keep the H2, but set it to font size preset XXL. In some cases, size XL might look better. Let's approach case by case.

Ok

> I suggest we replace the buttons for pill style paragraphs to better match the design.

@beafialho, I noticed that the links for the pills are underlined. Are you OK with this?

beafialho:
> I noticed that the links for the pills are underlined. Are you OK with this?


I'd prefer if they weren't, but I think we want to signal that those are supposed to be links. If we do, I'm ok with them being underlined.

juanfra:
> I'd prefer if they weren't, but I think we want to signal that those are supposed to be links. If we do, I'm ok with them being underlined.

Ok, I'll add your suggestion. So that we can move forward with this. We could also modify the pill style variation to remove the underline in links. But maybe we want to have some sort of audit first, to see if it has any other implications in other patterns.

beafialho:
> We could also modify the pill style variation to remove the underline in links. But maybe we want to have some sort of audit first, to see if it has any other implications in other patterns.

That would be great. I don't think there would be other implications in other patterns.

juanfra:
> That would be great. I don't think there would be other implications in other patterns.

Okay, we can work on that. In the meantime, I checked this one and saw that the new pill layout does not work well on mobile devices.

https://github.com/user-attachments/assets/57a41e56-2527-4225-9fd1-bab287b7b2f9

beafialho:
I'm not seeing that on my end. Does the row block containing them have "Allow to wrap to multiple lines"? That should fix it.

juanfra:
Good catch ✨

I've removed the underline from the links in the pill styles.

PR #265, opened 2024-09-10
Visual adjustments to Photo Blog templates to match design

beafialho:
**Description**

I tested these templates and noticed a few things that needed adjustments. I'd appreciate a quick review.

**Photo Single Post:**

- The layout wasn't looking as the intended design, so I adjusted blocks, spacings and sizes accordingly. This one is still missing the comments but I see that's being addressed in another PR.

Before:
https://github.com/user-attachments/assets/bcf60a5f-8ae8-40e4-b9a5-57081d71fe65

After:
https://github.com/user-attachments/assets/4cec2188-8335-44fe-8928-5be1459669cb

**Photo Archive Template:**

- Adjusted the Term Description font size in global styles

**Photo Search Results Template:**

- The search form had a size small, when it should have the default (Medium)

**Photo Home Template:**

- Added the pill style to "Stories"


carolinan:
One question: Is the locking and the custom block names on the archive and search intentional?

carolinan:
Oh never mind, those were not added as part of this PR, and can be discussed separately.

carolinan:
The author name does not have a hover style, it is always underlined, both in its default state and on hover. This may be unrelated to the PR, so I will merge this.

beafialho:
Could someone check if these changes are already reflected in trunk? I'm still seeing the old layout before the updates even after clearing cache.

PR #269, opened 2024-09-11
Add custom font sizes for headings.

### LINKED ISSUE #170: Add additional font sizes as a custom CSS property in theme.json

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/170

I'm adding custom font sizes for headings so that we can use the different heading sizes in other elements to respect the sizes in the design without neglecting accessibility concerns. 

The variables will be:
```
--wp--custom--font-size--heading-1
--wp--custom--font-size--heading-2
--wp--custom--font-size--heading-3
--wp--custom--font-size--heading-4
--wp--custom--font-size--heading-5
--wp--custom--font-size--heading-6
```

Tried to go with just `h1`, ..., `h6` as the values in the `theme.json` but then they were formatted as `--wp--custom--font-size--h-2` and I thought that having `heading` could result in a better devex. I'm happy to switch back to the other option if that's considered a better idea.

For the values, I went with the computed values using `clamp()`.

**Screenshots**

- N/A

**Testing Instructions**

- Open a page
- Check the CSS variables
- Confirm that you see the new variables with their values.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -926,7 +926,13 @@
 		},
 		"custom": {
 			"fontSize": {
-				"huge": "clamp(2.5rem, 4vw, 4.375rem)"
+				"huge": "clamp(2.5rem, 4vw, 4.375rem)",
+				"heading-1": "clamp(1.561rem, 1.2288rem + 1.661vw, 2.62rem)",
+				"heading-2": "clamp(1.146rem, 0.9408rem + 1.026vw, 1.8rem)",
+				"heading-3": "clamp(1.06rem, 0.878rem + 0.91vw, 1.64rem)",
+				"heading-4": "clamp(0.984rem, 0.8222rem + 0.809vw, 1.5rem)",
+				"heading-5": "clamp(0.875rem, 0.7574rem + 0.588vw, 1.25rem)",
+				"heading-6": "0.875rem"
 			}
 		},
 		"useRootPaddingAwareAlignments": true


carolinan:
> --wp--custom--font-size--h-2

That looks like bug 🐞 that should be reported, if an issue does not already exist.

carolinan:
These variables should be applied to elements > H1 etc  so that the sizes can be updated from one place.
Or did you think that should be a separate pull request?

juanfra:
> That looks like bug 🐞 that should be reported, if an issue does not already exist.

It's only happening when there's a combination of letters and numbers. I am not sure if it's intentional or if it's a bug. I can look and see if there's some more info in Gutenberg that can help us clarify that. And if it's a bug, report it.

> These variables should be applied to elements > H1 etc so that the sizes can be updated from one place. Or did you think that should be a separate pull request?

I believe that we should merge this one with the variables and then address these things in different PRs. There are many places where we'll need to replace values and I believe that by merging this one first and take care of the rest in other PRs will expedite things.

richtabor:
Why are we implementing custom font size variables instead of using the core method for registering font sizes? 

I don't think this is the correct approach here to be honest.

juanfra:
Hey @richtabor - Thanks for your input!

Carolina shared the reasoning behind this direction [here](https://github.com/WordPress/twentytwentyfive/pull/286#issuecomment-2347960322). It's a pretty good summary of what's happening.

To be honest, I'd prefer not having to do this, but we need to address the problem we're having. We need to find a balance that works well between the presets we have, design, and accessibility.

Let's continue the conversation [here](https://github.com/WordPress/twentytwentyfive/pull/286), and find the best approach.

PR #282, opened 2024-09-12
Remove all Categories: hidden in the pattern files

### LINKED ISSUE #259: Hidden patterns: Remove the categories
**Description**
Some of the hidden block patterns that have `Inserter: no` in the pattern file header also have a category listed. The category is "hidden".

This category should be removed, because "hidden" is not an existing category, and categories are not a required parameter
when registering patterns.

I have not found a reason for _why_ it is added, other than:
- It is added by the Create Block Theme plugin: [See conversation on the WordPress Slack (account needed).](https://wordpress.slack.com/archives/C015GUFFC00/p1725864890894449)
- It is included in Twenty Twenty-Four.

Update: There is a [pull request open](https://github.com/WordPress/create-block-theme/pull/718) to remove this from Create Block Theme.

KevinTaron:
Removes all locations where Categories: hidden was placed by the create block theme plugin.
Should solve the Issue #259

### DIFF
--- a/patterns/hidden-sidebar.php
+++ b/patterns/hidden-sidebar.php
@@ -2,7 +2,6 @@
 /**
  * Title: Sidebar
  * Slug: twentytwentyfive/hidden-sidebar
- * Categories: hidden
  * Inserter: no
  *
  * @package WordPress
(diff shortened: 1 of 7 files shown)


juanfra (approved):
Thank you! Looks good to me 🚀

PR #283, opened 2024-09-12
Add landing page for podcast and event

juanfra:
**Description**

Add landing pages for podcast and event.

Figma links:

- [Landing page: product/event](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=1-3413&t=kxoQKmyP8HzgJJSC-1)
- [Landing page: Podcast](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=1-3420&t=kxoQKmyP8HzgJJSC-1)

The only difference between the designs is that the "About the event" section for the "Landing podcast" has the default variation, and I don't know if that's possible. cc @carolinan 

**Screenshots**

https://github.com/user-attachments/assets/c4da94cf-4e65-4b05-9a53-094559ee2fba

**Testing Instructions**

1. Create a page.
2. Select the Landing page podcast and confirm that it's using the right patterns and in the same order
3. Repeat the same with the "Landing page for Event".

### DIFF
--- a/patterns/page-landing-event.php
+++ b/patterns/page-landing-event.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * Title: Landing page for Event
+ * Slug: twentytwentyfive/page-landing-event
+ * Categories: twentytwentyfive_page, featured
+ * Keywords: starter
+ * Block Types: core/post-content
+ * Post Types: page, wp_template
+ * Viewport width: 1400
+ * Description: .
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
+<!-- wp:pattern {"slug":"twentytwentyfive/hero-full-width-image"} /-->
+<!-- wp:pattern {"slug":"twentytwentyfive/heading-and-paragraph-with-image"} /-->
+<!-- wp:pattern {"slug":"twentytwentyfive/two-headings-and-a-list"} /-->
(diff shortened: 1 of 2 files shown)


beafialho:
Thank you @juanfra these are looking good to me!

I do see some minor strange paddings in `<!-- wp:pattern {"slug":"twentytwentyfive/heading-and-paragraph-with-image"} /-->` and `<!-- wp:pattern {"slug":"twentytwentyfive/two-headings-and-a-list"} /-->` that make it look misaligned with the other patterns. Should we address those in other PR?

carolinan:
Can you link to the design you are talking about?

carolinan:
No, there is no way for a template or pattern to change the style of the pattern.

juanfra:
> I do see some minor strange paddings in  and  that make it look misaligned with the other patterns. Should we address those in other PR?

@beafialho Those paddings are coming from the pattern implementation. To update them, then we'll have to update the patterns separately and it would be better to do it in separate PRs.

> Can you link tot he design you are talking about?

@carolinan I've updated the description with the Figma links.

beafialho:
> @beafialho Those paddings are coming from the pattern implementation. To update them, then we'll have to update the patterns separately and it would be better to do it in separate PRs.

Thank you, I'll address those in separate PRs then 👍 I don't think it's a blocker to merging this one though!

carolinan:
The patterns need a description, but let's merge this so that they can be tested.

There is no way, that I have found, to automatically select the `page-no-title` page template when the pattern is selected.

PR #284, opened 2024-09-12
Add news blog and archive query loop pattern

### LINKED ISSUE #275: News blog archive and News blog search results are referencing missing patterns.
**Description**
These two patterns include this markup:
`<!-- wp:block {"ref":2773} /-->` But this is a reference to a pattern on a single WordPress install.

Select and view these templates in the Site Editor to see the message "Block has been deleted...."

**Expected behavior**
I expect the template to reference a pattern for the post lists.

dballari:
**Description**

This PR solves the issue #275 

**Screenshots**

Captura de pantalla 2024-09-12 a las 13 19 30
Captura de pantalla 2024-09-12 a las 13 19 25

**Testing Instructions**
1. Temporarily modify archive and search templates by templates that only contain the news blog archive pattern and the news blog search result pattern respectively
2. This is achived by adding this code  and  respectively
4. perform a search and visit an archive page
5. Results must be like screenshots provided


KevinTaron:
Hey @dballari - I noticed the new footer-newsletter is not added to the theme.json - I think it should be added as a templatePart in theme.json

dballari:
Hi @KevinTaron yes you are right. Thanks a lot. I just added it.

carolinan:
Why is the template part needed instead of just using the correct pattern directly?

carolinan:
Ignore the question, I will assume it is for consistency with the other templates.

patterns/news-blog-query-loop.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/**
+ * Title: News blog query loop
+ * Slug: twentytwentyfive/news-blog-query-loop
+ * Template Types: archive
```
carolinan: This pattern that only has the query loop in it, is showing in the "Design" panel in the Site editor when the archive template is being edited. It is only meant to be part of another template.

The `Template Types` needs to be removed.
And then, because there will be no preview for this pattern left in the interface, the `Viewport width` can also be removed.

patterns/news-blog-query-loop.php
```diff
@@ -0,0 +1,64 @@
...
+<!-- wp:paragraph {"placeholder":"Add text or blocks that will display when a query returns no results."} -->
+<p>No posts were found.</p>
+<!-- /wp:paragraph -->
+<!-- /wp:query-no-results -->
+
+<!-- wp:spacer {"height":"var:preset|spacing|70"} -->
+<div style="height:var(--wp--preset--spacing--70)" aria-hidden="true" class="wp-block-spacer"></div>
+<!-- /wp:spacer --></div>
+<!-- /wp:query --></div>
+<!-- /wp:group -->
```
carolinan: The closing group here needs to be removed, it does not have a matching opening tag and block comment.
The Site Editor displays this as an HTML block:
image

dballari: you are right, I did not know this way of testing the template, thanks

patterns/news-blog-query-loop.php
dballari: yes they should be removed

dballari:
> Why is the template part needed instead of just using the correct pattern directly?

Hi @carolinan in the begining I did not think of it, I just did it for consistency, but now I'm thinking: we should be able to create new templates from the editor, right? not just from the code, and then for instance if we want a template for a certain category it will be nice to be able to add the footer newsletter template part as a template part, no just a pattern right? Anyway, there is something I'm missing here, I do not now how to retrieve the news blog templates from the editor ... I have a lot to learn here. I will be working on this until I fully understand how it works

carolinan:
I can help with that.

From the Site editor menu, -the dark grey sidebar, go to Templates.
To test the Archive, select archive, to test Home, select home, and so on.

While editing the template, you may first need to reset it: Because the editor does not actually show which pattern is currently used.
Open the settings sidebar.
Select the Template tab, not the block tab.
Next to the template name there is a menu with three dots. If there is anything to reset, it can be reset there.

Below the description is the Content panel. Skip this and open the next panel, Design.
All patterns for the selected template shows there.
Hover over them to see the pattern name, if needed (this may require Gutenberg, I don't remember for sure).
Click on the pattern you want to test.
This replaces the current content in the template with the selected pattern.
Save.
View the page on the front.

carolinan:
We have a design approval, the issue with the closing group block has been fixed, and all the templates have the footer updated, so lets merge this.

Issue #279, opened 2024-09-12
Right aligned blog with vertical heading: Make sure there is only one H1 heading


carolinan:
**Description**

On the right aligned blog with vertical heading, the site title is set to the H1 heading level.
The site title should not be used as the main H1 of the page, since it describes the site, not the content on the current page.
On some of the templates, there are currently two H1 headings.

Tasks:
- In the templates for this design, set the site title to a paragraph but make sure that it uses all the correct typography settings. The **look** of the site title mustnot change.
- Make sure each template has a single H1. In some of the templates, the H1 already exists, the others need to be updated.

carolinan:
Actually, the "header" in these templates need to be added as a template part, otherwise, the navigation will not be the same  for the template set. The user may accidentally add one menu for the home and a different one for the single.

KevinTaron:
> Actually, the "header" in these templates need to be added as a template part, otherwise, the navigation will not be the same for the template set. The user may accidentally add one menu for the home and a different one for the single.

Is that even possible? In the "header"-templatepart we have a row instead of a stack and the Navigation is not Overlay-Menu: "Always"

carolinan:
It needs to be a new template part.

KevinTaron:
added a pull request for this #285

PR #285, opened 2024-09-12
Create the header template part for the right aligned blog with vertical heading

KevinTaron:
**Description**
Regarding Issue #279 - the vertical header was outsourced to a template part and registered in the theme.json.
The pattern for the vertical header was inserted and the title was changed to level 0.

The new template part was inserted at all places where the original vertical header was used. 
Also all there is now only one h1 on every template. But for the home template "vertical-header-aligned-home-template.php" I have added a h1 with "Blog" like the original home.html - because the blog posting shouldn't be the h1 and there was no other good option to set the h1. 
In the archive it's the archive title, on the single pages and posts its the title and in the search results its the query title.

There shouldn't be visible changes as I tested, except the "Blog" title in the home-template.php

** Screenshots ** 

Screenshot of Blogtitle and figma design
image
image

**Video**

[output.webm](https://github.com/user-attachments/assets/389cb42f-d77e-46c4-a0ec-9994a1c98bc1)

**Testing Instructions**

1. Active the theme and add the template to the pages (home-template to home, single-page to page etc.)
2. Check if everything is working correctly and looks like before and the figma design.

### DIFF
--- a/parts/vertical-header.html
+++ b/parts/vertical-header.html
@@ -1,0 +1,1 @@
+<!-- wp:pattern {"slug":"twentytwentyfive/vertical-header"} /-->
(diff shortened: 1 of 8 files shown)


beafialho:
Thank you @KevinTaron!

Could you please add a similar spacer (`preset Regular`) below "Blog"?

Captura de ecrã 2024-09-12, às 16 32 56

KevinTaron:
Hope you meant this spacer @beafialho

carolinan:
I have merged this in a separate PR because of the merge conflicts. I have added the props from both PR's.

Issue #186, opened 2024-09-03
Accessibility: Replace H1's in non-page patterns


carolinan:
**Description**

On archives, index, blog/home, and search, the heading above the results should be the only H1.
On single pages and posts, the post title should be the only H1.

When the user adds a pattern that is intended to be one part of the page, that pattern should not contain an H1,
because then by default, there are two H1's, which makes it more difficult for some users to determine what the most important content on the page is.

Landing page patterns, that are intended to be the only pattern on the page. must contain an H1, and only one H1.
(A landing page can be one pattern that is combined of several other patterns)
_Update: These landing pages will need to rely on the user making the right choice, either use the post title, add an H1, or update the levels._

---
Task;:
Review the use of H1's in the current patterns.
Replace H1's with appropriate levels but make sure that the look is the same as before, including: font size, letter spacing, weight etc.

carolinan:
As @richtabor correctly pointed out, even if the landing page patterns were designed to be used on the page template that has no title, the default is to use the template that has the title.

The H1 heading should be replaced on the landing page patterns.

PR #286, opened 2024-09-12
Fix heading sizes, to use custom sizes and match designs.

juanfra:
**Description**

Now that we have the [custom heading sizes](https://github.com/WordPress/twentytwentyfive/pull/269) we need to fix the font sizes and HTML elements for most of the patterns, to match the designs.

Partial for https://github.com/WordPress/twentytwentyfive/issues/186

Patterns:

- Clients section
- Pricing two columns
- Event schedule
- Events – List with Call to Action
- Grid with videos
- Grid with categories
- Hero Centered Heading
- Hero podcast
- Short heading and paragraph and image on the left
- Services - Team photos
- Two Headings and a List
- Heading and paragraph with image on the right
- Services - Subscriber only section
- Events three columns with event images and titles

**Screenshots**

N/A

**Testing Instructions**

- Create a page.
- Add the following patterns modified in this PR.
- Confirm that the headings are using the proper HTML elements (to improve a11y) and the same sizes as the designs.

### DIFF
--- a/patterns/clients-section.php
+++ b/patterns/clients-section.php
@@ -15,8 +15,8 @@
 <div class="wp-block-group alignfull is-style-section-5" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--80);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--80);padding-left:var(--wp--preset--spacing--50)">
 	<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"top":"0","bottom":"0","left":"0","right":"0"}}},"layout":{"type":"default"}} -->
 	<div class="wp-block-group alignwide" style="padding-top:0;padding-right:0;padding-bottom:0;padding-left:0">
-		<!-- wp:heading {"textAlign":"center","level":3} -->
-		<h3 class="wp-block-heading has-text-align-center">The Stories Podcast is sponsored by</h3>
+		<!-- wp:heading {"textAlign":"center"} -->
+		<h2 class="wp-block-heading has-text-align-center">The Stories Podcast is sponsored by</h2>
 		<!-- /wp:heading -->
 
 		<!-- wp:group {"align":"wide","style":{"spacing":{"blockGap":"var:preset|spacing|20","padding":{"right":"0","left":"0","top":"var:preset|spacing|50","bottom":"0"}}},"layout":{"type":"flex","flexWrap":"wrap","justifyContent":"center"}} -->
--- a/patterns/cta-pricing.php
+++ b/patterns/cta-pricing.php
@@ -16,8 +16,8 @@
 <div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0;padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)">
 	<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"bottom":"var:preset|spacing|50"}}},"layout":{"type":"flex","orientation":"vertical","justifyContent":"center"}} -->
 	<div class="wp-block-group alignwide" style="padding-bottom:var(--wp--preset--spacing--50)">
-		<!-- wp:heading {"textAlign":"center","align":"wide"} -->
-		<h2 class="wp-block-heading alignwide has-text-align-center">Pricing</h2>
+		<!-- wp:heading {"textAlign":"center","align":"wide","style":{"typography":{"fontSize":"var(--wp--custom--font-size--heading-1)"}}} -->
+		<h2 class="wp-block-heading alignwide has-text-align-center" style="font-size:var(--wp--custom--font-size--heading-1);">Pricing</h2>
 		<!-- /wp:heading -->
 
 		<!-- wp:paragraph {"align":"center"} -->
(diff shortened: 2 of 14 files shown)


richtabor:
I'm not particularly confident in this approach. What is the reasoning behind this again? Can we not use font sizes in their standard implementation?

carolinan:
I initially recommended using five font sizes where the heading sizes matches the presets. 
The compromises was six sizes. Then later the sixth size was removed because the "t-shirt size" control only shows five.

Additional font sizes have been added during reviews, resulting in some elements having one-off sizes in the patterns, which makes the style variations / typography presets difficult to use.
One-off sizes is not a problem if they are limited to a few, like in a single pattern with a gigantic text.
It does become a problem when the sizes are needed to make an H2 look like an H1.

Last Friday I suggested that we either use the selectable font sizes for the headings, or add the heading sizes as custom CSS properties in theme.json. 

There will always be compromises. I don't want to put words in @beafialho's mouth, so to say, but there has already been a number of compromises including the accessibility related changes, and features that are missing in the editor which has meant that some parts of the design has not been possible. It is important that @beafialho can still feel that the theme lives up to the vision and design.

beafialho:
Could someone please explain to me what visual changes this PR will imply? What do I need to look when testing? Were there changes to my previous PR #231 again?

beafialho:
Thank you, I appreciate your attention in wanting to maintain the original design hierarchy.

I tested this and my take, looking at the patterns without these custom sizes, I don't think we absolutely need this fix. 

The change I made to the heading sizes in #231 already accounted that some of the patterns might look slightly different, but I don't think they look bad, especially because we have been using the correct heading levels for accessibility and applying some of the `L / XL / XXL` font size presets _when necessary_ to improve their look.

That said however, if we do want to use this approach, the patterns look good to me as well 👍

carolinan:
The main issue is not with the look of these pattern, but that the heading sizes are not included in the font size presets.
H1 could have matched the largest font size preset, H6 the smallest but with uppercase, and so on.
Then the custom CSS variables would not be needed, and the user would be able to select from all the sizes.

richtabor:
```
"custom": {
			"fontSize": {
				"huge": "clamp(2.5rem, 4vw, 4.375rem)",
				"heading-1": "clamp(1.561rem, 1.2288rem + 1.661vw, 2.62rem)",
				"heading-2": "clamp(1.146rem, 0.9408rem + 1.026vw, 1.8rem)",
				"heading-3": "clamp(1.06rem, 0.878rem + 0.91vw, 1.64rem)",
				"heading-4": "clamp(0.984rem, 0.8222rem + 0.809vw, 1.5rem)",
				"heading-5": "clamp(0.875rem, 0.7574rem + 0.588vw, 1.25rem)",
				"heading-6": "0.875rem"
			}
		},
```

This will just confuse people. Why the theme is defining custom font sizes on top of the standard implementation. We should not do this. 

> It does become a problem when the sizes are needed to make an H2 look like an H1.

Why not use a block style variation for text styles?

carolinan:
> Why not use a block style variation for text styles? 

That is possible, and it may be a helpful addition to users 👍 But it wouldn't actually solve the problem with the large number of font sizes. The value still needs to be defined, and defined in one place, so that it can be updated from one place if needed.

carolinan:
I don't think a style variation should be used **only** to change a size. That would seem incredibly redundant to me, when there is a size option that could do the same, if only the presets and the heading sizes were the same.

juanfra:
To give more context about the reasoning behind this:

- Most pattern designs have a main title using the size of an H1, and for a11y reasons, we can't use H1 on most. Using an H2, would make these titles more accessible, but the size is too small compared to the designs.
- We have five font sizes (S, M, L, XL, XXL) that do not match the Headings. If we set an H2 with XL, it looks too small, and XXL is too large.

Thinking of a11y as one of the top priorities, the patterns' main titles should have an H2 element, but we needed to figure out how to handle the size (for the reasons described above) so that the patterns looked like they did in Figma.

That's why Carolina suggested using custom variables with the heading sizes that we could use in the code. This PR uses the sizes we introduced, so the H2s have the sizing of the Figma designs. From what I understand, the custom variables are not seen in the UI, and it could be invisible to many users.

I believe that if we don't want to go down this road, we should consider:

- Making a pass on patterns and considering updating them, so that the titles use one of the sizes we have in the presets. In theory, the size should be smaller than H1, so that when the patterns are inserted in the context of a page with title, the hierarchy is respected.
- OR review the heading sizes, but that was already checked [here](https://github.com/WordPress/twentytwentyfive/pull/231), so reverting the heading sizes will put us in the same situation with the headings that Bea fixed. So, I personally believe this is a no-go.

In my opinion, our main problem is that the size of the H2 doesn't match (or closely match) the presets we have. If we figure that out, we should be good without this. Because the problem is that we need to use H2's for a11y purposes, and that doesn't match the design.

I believe that having presets for headings only changing the size could be confusing for users, as they'll have three places competing for the same functionality. They'll be able to change the size from the level selector (H1, H2, .., H6), the sizes in the typography section, and the variation.

beafialho:
> If we set an H2 with XL, it looks too small, and XXL is too large.

Just to make sure it's clear, as a compromise, this should be fine. I've been using these two presets in templates when the Heading 2 size is too small.

I strongly advocate that we do not change the heading sizes one more time. I have defined them so they look perfect in the default template and across the others. If we keep on changing them, we risk having things looking [like this](https://github.com/WordPress/twentytwentyfive/pull/231) again, which I'd really like to avoid.

beafialho:
> Making a pass on patterns and considering updating them, so that the titles use one of the sizes we have in the presets. In theory, the size should be smaller than H1, so that when the patterns are inserted in the context of a page with title, the hierarchy is respected.

This seems like the way to go to me 👍

carolinan:
But why cant the largest preset match the H1 heading size?
Why do these need to be different?

beafialho:
When the design was done, I had defined plenty more font sizes than the current ones, especially because of the patterns. We've been removing more and more sizes as the theme gets built. I think we've compromised enough and reached a good number. Why do they need to be the same size?

carolinan:
The user can only select from 5 right now.
So there is no way for them to make their own texts match the rest of the theme.
Even if they had 12 sizes to choose from, it would be impossible for them to pick the correct and matching size.

beafialho:
Sorry, but I don't think I'm understand the problem. Could you exemplify with a recording or screenshot?

carolinan:
> Sorry, but I don't think I'm understand the problem. Could you exemplify with a recording or screenshot?

If a user adds a pattern where there is a heading H2 that has the custom font size -H1, there is no interface, no setting, for the user to also add an H2 heading with the custom font size H1. They can't select that size for their H2.
So they can't create a custom section that matches the same design. Unless they realize they can just copy the pattern, of course.

https://github.com/WordPress/twentytwentyfive/issues/305 can resolve that, but it can also be overwhelming if all sizes are added there.

beafialho:
> If a user adds a pattern where there is a heading H2 that has the custom font size -H1

So yet another reason not to add custom heading sizes, right? On the other hand, if a user adds a pattern where there is a heading 2 with the XL font size preset, they can remove that preset or pick another from the typography panel.

carolinan:
The XL font size is in the option though. So that is not a problem.
The H1 heading font size is not.

-----
Do I understand correctly that the suggestion is to close this PR and not do these changes?

beafialho:
I'm not suggesting closing because I'm not sure what's the best solution.

Mostly trying to understand what the issue is and expressing my concern: that the font sizes have already been changed a few times, I've already gone through the heading sizes to get them to look _just right_, both in the posts, as well as in the patterns (using the correct headings and applying the presets as necessary).

If we want to add these custom sizes, I am fine with that as long as we do not move backwards on how things look across the theme.

richtabor:
I'm working on another variation of this.

juanfra:
Closing in favor of https://github.com/WordPress/twentytwentyfive/pull/422

Issue #17, opened 2024-07-15
Post format archive templates


carolinan:
It has probably not gone unnoticed that there is a request to[ add support for post formats](https://make.wordpress.org/core/2024/07/08/what-would-you-like-to-see-in-the-next-default-wordpress-theme/#comment-46916) to block themes in general and Twenty Twenty-Five.

Currently, there are only two things that the **theme** can do to support post formats:
- Register them using the theme support (Post formats are not enabled by default) `add_theme_support( 'post-formats')`
- Include HTML templates for the post format archives. 

From the theme developer handbook:
>In the case of [post formats](https://developer.wordpress.org/themes/functionality/post-formats/), the taxonomy is ‘post_format’ and the terms are ‘post-format-{format}. i.e. taxonomy-post_format-post-format-link.php for the link post format.

https://developer.wordpress.org/themes/basics/template-hierarchy/#custom-taxonomies

justintadlock:
I did quite a bit of exploration for post formats here: https://github.com/x3p0-dev/x3p0-ideas/issues/17

It's probably a bit more involved than we'd want to include a default theme, but just sharing in case any of it's useful.

beafialho:
What else is needed from design here? There are "post format" templates mocked up [here](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3259-4046&t=ZfLLXJppTnoBfVEI-4).

carolinan:
Then I will remove the needs design label.

carolinan:
@beafialho The theme can include archive templates for post formats. But not templates for single post formats. It is not part of the template hierarchy.

carolinan:
Hm, actually that does not work either. When the query loop is set to inherit the query from the template, the editor shows all posts, but the front shows the correct post with the expected format.

Branch for testing: https://github.com/WordPress/twentytwentyfive/tree/add/post-formats
Create a new post and assign it the link format. Save.
Go to the front of the site and add type/link : yoursite.domain/type/link.
Confirm that the archive shows the post you created.
In the admin toolbar, select "Edit site" to open the template for the link format.
The query block, which is supposed to inherit the parameters from the template, shows all posts.

jasmussen:
I can confirm your steps.

Post:
![Screenshot 2024-09-24 at 13 03 43](https://github.com/user-attachments/assets/9d31d5cb-a6e4-4256-9fa3-ddfbf3dab044)

[...]

PR #295, opened 2024-09-13
Add theme support for post formats

carolinan:
**Description**
I have re-opened this PR after the discussion in https://github.com/WordPress/twentytwentyfive/issues/17
If we want the theme to have partial support for post formats in 6.7, it needs to be added to the theme before Beta 1, since it is an enhancement. Then if it does not work well enough, it can be removed before the string freeze.

This PR:
- Registers theme support for the formats that are listed in Figma: gallery, quote, link, audio, video
- Adds the HTML archive template for the "link" post format.

**Testing Instructions**

This can be tested with the latest version of Gutenberg active.

1. Set the permalinks on the test installation to anything but plain which does not seem to work.
2. Create one or more posts and assign the "link" post format.
3. View the archive on the front: Add `type/link/`after your test domain in the browser address bar.
9. Confirm that the query loop shows the correct posts.


functions.php
```diff
@@ -9,6 +9,17 @@
  * @since Twenty Twenty-Five 1.0
  */
 
+/**
+ * Add theme support for post formats.
+ *
+ * @since Twenty Twenty-Five 1.0
+ * @return void
+ */
+function twentytwentyfive_post_format_setup() {
```
juanfra: Let's wrap this in an `if ( ! function_exists(...` like the other functions in the file.

templates/taxonomy-post_format-post-format-link.html
```diff
@@ -0,0 +1,41 @@
+<!-- wp:template-part {"slug":"header","area":"header","tagName":"header"} /-->
```
juanfra:
```suggestion
<!-- wp:template-part {"slug":"header"} /-->
```

templates/taxonomy-post_format-post-format-link.html
```diff
@@ -0,0 +1,41 @@
...
+				<!-- wp:query-pagination-next /-->
+			<!-- /wp:query-pagination -->
+		</div>
+		<!-- /wp:group -->
+	</div>
+	<!-- /wp:query -->
+</main>
+<!-- /wp:group -->
+
+<!-- wp:template-part {"slug":"footer","tagName":"footer"} /-->
```
juanfra:
```suggestion
<!-- wp:template-part {"slug":"footer"} /-->
```

juanfra (commented):
Thanks Carolina, so the idea is to squeeze it in before beta 1 in case we can make it for 6.7. Right?

templates/taxonomy-post_format-post-format-link.html
carolinan: I don't think is going to be the final template. 🤷‍♀️

carolinan:
> Thanks Carolina, so the idea is to squeeze it in before beta 1 in case we can make it for 6.7. Right?

I do not believe there is enough support or developers available, but, yes.

jasmussen:
Appreciate this PR.

Do we need additional templates or is this ready to land?

carolinan:
There should be archive templates for each supported format.

And possibly for the singles. The good thing is there are Figma designs for the singles.

jasmussen:
I'll see if I can do anything to help here.

jasmussen:
Following up, I created https://github.com/WordPress/twentytwentyfive/pull/441 with a summary of some learnings. 

One path forward I can see is: we register these formats, but we don't add templates for them. And we create a few patterns for the slightly more elaborate formats, Link and Audio. Let me know your thoughts.

carolinan:
- I removed the HTML template
- Wrapped the theme_support in function exists
- Added the image post format - We did not discuss this, but I think it will be requested by users.

jasmussen (approved):
This feels worth trying, and it will be easier to rewind if we need to, than add at a later time.

PR #298, opened 2024-09-13
All patterns that are not hidden needs a description #290

### LINKED ISSUE #290: All patterns that are not hidden needs a description
**Description**

Block patterns that are not hidden from the inserter needs a description, a short text that describes the pattern to users who are not able to see the pattern preview.
This description is visually hidden in the interface but available for screen-readers.
It has the same purpose as an alternative text on an image, and helps users choose which pattern to insert.

### Don't:
```
 * Title: Portfolio Homepage
 * Description: A portfolio homepage pattern.
```
This description does not bring any additional value.

` * Description: .`

This description will be announced, but does not bring any value.

` Description: Displays a list of posts with title and date`
"A list of posts with titles and dates" would be enough, and shorter.

### Do
` * Description: A grid section with photos and a link to an Instagram profile.`

### Maybe?
` * Description: A subscriber-only section highlighting exclusive services and offerings.`
It is not wrong, but does not explain what blocks are used or what it looks like.

noruzzamans:
I have updated the descriptions for 5 patterns. Please review them and let me know if I followed the correct format. If so, I will update the rest of these. Thanks!

Closes: #290

### DIFF
--- a/patterns/post-with-left-aligned-content.php
+++ b/patterns/post-with-left-aligned-content.php
@@ -14,72 +14,100 @@
 ?>
 <!-- wp:template-part {"slug":"header","tagName":"header","area":"header"} /-->
 
-<!-- wp:group {"align":"wide","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"default"}} -->
-<div class="wp-block-group alignwide"><!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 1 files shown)


carolinan:
Thank you for working on this.

All these descriptions are too long: 

The user has a list of patterns to navigate through. They may be searching for a specific pattern, and they may already know which pattern they want to insert.
So needing to listen to lengthy descriptions is not helpful, it can instead be frustrating.

noruzzamans:
> Thank you for working on this.
> 
> All these descriptions are too long:
> 
> The user has a list of patterns to navigate through. They may be searching for a specific pattern, and they may already know which pattern they want to insert. So needing to listen to lengthy descriptions is not helpful, it can instead be frustrating.

@carolinan Okay, I will make it short and meaningful. Thanks!

noruzzamans:
@carolinan I have updated this five-pattern description. Could you please review it again? Thanks!

Issue #300, opened 2024-09-13
Don't include left and right padding in fullwidth constrained patterns


richtabor:
WordPress will add the outer padding value of the site instead. If we fix these values, then the control in styles > layout will not change the outer padding. 

Here's what happens if the values are fixed: 

https://github.com/user-attachments/assets/94d91ad2-1375-46d2-99a9-e2ce8d2c633c

And here's when the fixed left/right padding values are removed: 

https://github.com/user-attachments/assets/bb43da31-d922-479b-a64c-74f65e3a0e16

carolinan:
This is already a known issue for some patterns, as discussed in the open pull requests.

carolinan:
Please motivate in detail why the patterns should change based on the _global spacing used to manage the spacing on the full pages._
Without a motivation it is difficult for other contributors to make informed decisions.

richtabor:
> If we fix these values, then the control in styles > layout will not change the outer padding.

I did include motivation. If we provide fixed values, then the site's outer padding value will not apply. You can change this control value, but have almost no effect (or worse, a variable effect).

If we want people to be able to change the left/right padding of their sites (which we should, as core supports it) then we should make it possible.

juanfra:
@richtabor can you please confirm if this one was fixed [here](https://github.com/WordPress/twentytwentyfive/pull/397)?

PR #301, opened 2024-09-13
Remove left/right padding on heading-and-paragraph-with-image pattern

richtabor:
Part of #300

CleanShot 2024-09-13 at 16 22 40

### DIFF
--- a/patterns/heading-and-paragraph-with-image.php
+++ b/patterns/heading-and-paragraph-with-image.php
@@ -11,9 +11,9 @@
  */
 
 ?>
-<!-- wp:group {"align":"full","className":"is-style-section-4","style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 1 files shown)


carolinan:
Please motivate in detail why the top and bottom margin 0 was removed.
Without a motivation, it is difficult for other contributors to make informed decisions.

carolinan:
I think this is what you meant with the _margins_? https://github.com/WordPress/twentytwentyfive/issues/328

richtabor:
> Please motivate in detail why the top and bottom margin 0 was removed.
> Without a motivation, it is difficult for other contributors to make informed decisions.

Yes, otherwise you can't place them next to paragraphs etc. We can potentially revisit, but for consistency let's roll without margins. 

> I think the best would be to remove the margin setting from the patterns, and then on the full page patterns, where these full width patterns are combined into one cohesive layout, wrap them inside an extra group with block spacing set to 0.

jasmussen (approved):
Thanks for clarifying the motivation, it makes sense to me. Unless @carolinan objects, we can go with this.

jasmussen:
It needs a rebase.

carolinan:
I believe these changes are part of https://github.com/WordPress/twentytwentyfive/pull/397/files#diff-73ae8b45ec4bf5e15c354067badb29af0c74955c819e869635dde3f7e4d3637aR13 and that this can be closed?

richtabor:
Closing in favor of https://github.com/WordPress/twentytwentyfive/pull/397

Issue #294, opened 2024-09-13
Feedback: The line length on the home and archives of the right aligned blog feels too long


carolinan:
**Description**

On my screen which is set to 1920x1080, the number of characters in the text content on the home page, search and archive of the "right-aligned blog" is 161 per line.

Even though it only shows the excerpt, I do need to move my eyes to scan the content.
This is made worse for me at least since the excerpt do not have natural line breaks and the text becomes difficult to scan without deep concentration.

Comparing with the footer, I am not sure that the content is set to the correct width?

![66 local__paged=2](https://github.com/user-attachments/assets/eff23dd2-4b6f-44eb-ab29-7f5a89d58434)

PR #316, opened 2024-09-14
Feedback: The line length on the home and archives of the right align…

gohelkunjan:
…ed blog feels too long Fixed

**Description**
#294 

**Screenshots**
image

**Testing Instructions**


carolinan:
@gohelkunjan I think there is some confusion here, the template that you are editing here is not the template with the long line length. 

Until Beatriz is able to reply to the issue, we can not even know if it is intentional or if it is something that needs to be changed.

gohelkunjan:
Okay, Thank for update.

Issue #278, opened 2024-09-12
Add missing <main> elements to template patterns


carolinan:
**Description**

These templates are missing the `<main>` element and it needs to be added:

- [x] news-blog-archive-template.php, News blog archive
- [x] news-blog-home-template.php,  News blog home
- [x] news-blog-single-post-with-sidebar-template.php News blog single post with sidebar
- [x] news-blog-with-featured-posts-grid-template.php, News blog with featured posts grid
- [x] news-blog-with-sidebar-template.php News blog with sidebar
- [x] news-blog-search-results-template.php
- [x] offset-post-no-featured-image-template.php Offset post without featured image
- [x] vertical-header-right-aligned-archive-template, Archive for the right aligned blog
- [x] vertical-header-right-aligned-home-template.php Homepage for right aligned blog
- [x] vertical-header-right-aligned-page-template.php  Page template for the right aligned blog
- [x] post-with-left-aligned-content.php
- [x] vertical-header-right-aligned-posts.php
- [x] vertical-header-right-aligned-search-results-template.php

**Expected behavior**

All pages (templates) has one `<main>` element.

All templates must include main, both because all content needs to be inside a landmark, but also because in block themes, the automated skip link requires it.
https://make.wordpress.org/themes/handbook/review/accessibility/required/#aria-landmark-roles

carolinan:
When working on these, please submit smaller pull requests since they will be easier to review and test.

noruzzamans:
@carolinan I am interested in working on these issues. Could you please share an example of how to resolve them? Thanks

KevinTaron:
I've added a pull request for all templates with named veritcal-header-right...
#280 

@noruzzamans maybe you can look into my pull request — that's an example how to do it. 
Probably you should wait and see if the maintainers accept the pull request as it is. ;)

carolinan:
> @carolinan I am interested in working on these issues. Could you please share an example of how to resolve them? Thanks

Some patterns needs a new block inserted, that will be used as the main, others need the HTML tag on existing groups to be updated.

When changing a group block's HTML tag from `<div>`, to `<main>`, both the block comment (the JSON-style attributes) and the  tag on the block itself needs to be updated.

From
```
<!-- wp:group {"layout":{"type":"constrained"}} -->
<div class="wp-block-group"></div>
<!-- /wp:group -->
```
To

```
<!-- wp:group {"tagName":"main","layout":{"type":"constrained"}} -->
[...]

PR #317, opened 2024-09-14
Add <main> to "Post with left-aligned content".

carolinan:
**Description**
Adds the `<main> `HTML tag to the pattern. All templates must include a main, both because all content needs to be inside a landmark, but also because in block themes, the automated skip link requires it.
Indents the blocks in the file to improve the readability.

Partial for https://github.com/WordPress/twentytwentyfive/issues/278

**Screenshots**

Before:
left-aligned-before

After:

left-aligned-after

**Testing Instructions**

Apply the PR and go to Appearance > Editor > Templates.
Select the single post template.
In the settings sidebar, open the Template tab.
Open the Design panel and select the "Post with left aligned content.
Open the list view.
Select the group block that is between the header and footer.
In the block settings sidebar, open the Advanced panel.
Confirm that main is selected in the HTML element option.
Save.
View a single post on the front.
Confirm that there are no issues with the layout, the template should look identical before and after adding the main.

### DIFF
--- a/patterns/post-with-left-aligned-content.php
+++ b/patterns/post-with-left-aligned-content.php
@@ -14,8 +14,8 @@
 ?>
 <!-- wp:template-part {"slug":"header","tagName":"header","area":"header"} /-->
 
-	<!-- wp:group {"align":"wide","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"default"}} -->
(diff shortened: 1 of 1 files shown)


juanfra (approved):
Looks good to me, thanks Carolina.

PR #320, opened 2024-09-15
Default single post template: Move the category below the title

### LINKED ISSUE #291: Default single post: The category above the post title needs to be moved
**Description**

In the default single post template / pattern, there is a category above the main heading.

-Please await information about the new position for the category before starting to work on this.

carolinan:
**Description**
This PR updates the default template for the single post, and closes https://github.com/WordPress/twentytwentyfive/issues/291.

It moves the category below the featured image, to the same line as "Written by [authorname]",
because positioning it before the title is not as accessible as positioning it below the title.

The new text is: `Written by Author name in category, category 2, category 2`.
The text that was already in the template was not translation ready, so I created a new hidden pattern with the slug
`hidden-written-by`.

This text has two text colors: It uses `primary` for the plain text, and inherits the default link color, `contrast`,  for the links.
There is one inconsistency in this PR compared to the proposed design in https://github.com/WordPress/twentytwentyfive/issues/291, and that is that the category links are not underlined by default.
And that is because the underline is hidden by default in theme.json. Changing this would create _other_ miss-matches between the templates and the design elsewhere.

The word "in" is a paragraph. This can be changed to be the prefix of the terms block if that is preferred.

**Screenshots**
Editor:

The Written by text and the category on the single post template in the Site Editor

Front, only the post meta:
single-post-front--post-meta

**Testing Instructions**
View the default single template in the editor and front and confirm if it matches the design.


beafialho:
I don't see the sentence with the post author and categories in the template, neither on the front end.

carolinan:
Because the text needs to be translatable, I moved it to a pattern. The pattern was then added to the template. I will double check ⏲️

carolinan:
It is working for me. Perhaps if you try to reset the template? Maybe there are other changes saved.

juanfra (approved):
Looks good to me, works well.

https://github.com/user-attachments/assets/7ae9dd0a-f85d-4a60-8bbc-204992b32293

beafialho:
Oh, it must be something in my environment. Let me try again.

patterns/hidden-written-by.php
```diff
@@ -0,0 +1,24 @@
...
+ * Inserter: no
+ *
+ * @package    WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since      Twenty Twenty-Five 1.0
+ */
+
+?>
+<!-- wp:group {"style":{"spacing":{"blockGap":"5px","margin":{"bottom":"var:preset|spacing|60"}}},"fontSize":"small","layout":{"type":"flex","flexWrap":"nowrap"}} -->
+<div class="wp-block-group has-small-font-size" style="margin-bottom:var(--wp--preset--spacing--60)">
```
beafialho, line 14: Would there be an issue if we set this spacing to `0,2em`? It looks a bit sharper than with `5px`.

beafialho (commented):
It's working now for me :)

I left one comment about the spacing of the post author and categories pattern. Other than that, it's looking good to me.

patterns/hidden-written-by.php
beafialho: Here's the difference:

https://github.com/user-attachments/assets/bcd1ba0f-734b-48dd-a8ec-0c22b751a277

carolinan: Done!

PR #321, opened 2024-09-15
Text-only single post: Move the category below the title

### LINKED ISSUE #293: Text-only blog single post: The category above the post title needs to be moved
**Description**

In the text-only blog single post template / pattern, there is a category above the main heading.

-Please await information about the new position for the category before starting to work on this.

carolinan:
**Description**
This PR adjusts the single post pattern for the text-only blog.
It moves the category below the post title and closes https://github.com/WordPress/twentytwentyfive/issues/293.
The reason for this is that displaying the category above the main heading is less accessible than displaying it below.

Changes made:
- I moved the post terms block below the post title.
- I removed the `preset-50` bottom margin on the post title so that the category could be displayed closer to the title.
- I added a `preset-50` spacer block between the post terms and the post content.

There is one inconsistency compared to the proposed design: the categories are not underlined by default.
This is because the post-terms block has the underline hidden by default, and showing on `:hover.`

**Testing Instructions**

Please see the new design in the linked issue, then enable the text-only design for the single post and confirm if the two results match:

Apply the PR.
Go to Appearance > Editor
Select Templates.
Open the Single post template
In the settings sidebar, open the Template tab.
Open the Design panel
Select the "Text-only blog, single post" pattern.

**Screenshot**
The text-only single post, on the front:

A single post with the text-only template, showing the category below the title, and the tags below the content.

### DIFF
--- a/patterns/text-only-blog-single-post-template.php
+++ b/patterns/text-only-blog-single-post-template.php
@@ -18,9 +18,11 @@
 <main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
 	<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
 	<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)">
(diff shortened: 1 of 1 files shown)


beafialho:
Thanks, it's looking good.

Could you remove the `XXL font size` preset to the Title? I missed this one when I removed it in the other templates.

carolinan:
Yes :)

PR #322, opened 2024-09-15
News blog single post: Move the post date below the post title

### LINKED ISSUE #292: News blog single post: The date above the post title needs to be moved
**Description**

In the news blog single post template / pattern, there is a date above the main heading.

-Please await information about the new position for the date before starting to work on this.

carolinan:
**Description**
This PR adjusts the pattern for the single post in the "News blog" set, and closes https://github.com/WordPress/twentytwentyfive/issues/292.

Changes made:
The tags where moved below the post content, and the right aligned text setting was removed.
A spacer block with `preset-40` was added between the content and the tags.

The post author avatar and name was moved to the right inside the row that is between the post title and the featured image.

The post date and categories were added on the left side of the row that is between the post title and the featured image.
The categories were set to uppercase, and are separated with a comma.

A divider was added between the post date and categories. This is the same "middle dot" that is used in the "News blog with sidebar" and "Post with left aligned content",
I was not sure which spacing to use: I set the block spacing on the row to 4px, which matches "Post with left aligned content".

When testing the the responsive view, I found that if there is more than one category, the date, categories, and author did not wrap well, so I changed "nowrap" on the outer row to "wrap".
On small screens, the date and categories are now on one line, and the author avatar and author name is on a separate line below.

The file was indented, and spaces were changed to tabs.

**Testing Instructions**
Please see the proposed design in the linked issue, then compare it with this PR:

Go to Appearance > Editor.
Select Templates.
Open the single post template.
In the settings sidebar, select the Template tab.
In the Design panel, select the pattern called "News blog single post with sidebar"
Save.

**Screenshots**
After, on the front (excuse the size, I reduced it to 70% but it is still 4Mb!)
news-blog-front


beafialho:
Thank you @carolinan, this is looking good.

The tags are placed correctly before the sidebar comes in on mobile (I believe that answers your questions in #292)

Only one small detail to be changed: the categories should have `1.4px` letter spacing.

PR #329, opened 2024-09-16
Remove the default palette, gradient and duotone

### LINKED ISSUE #308: Remove default presets for colors, gradients, and duotone filters
We should remove the default presets for colors, gradients and duotones. They don't add to the experience; likely get in the way. 

CleanShot 2024-09-13 at 17 02 16
CleanShot 2024-09-13 at 17 02 32
CleanShot 2024-09-13 at 17 03 01

carolinan:
**Description**
Sets the default color palette, gradients and duotones to false in theme.json, so that they do not distract from the theme colors.
Closes https://github.com/WordPress/twentytwentyfive/issues/308

**Testing Instructions**
Insert a block that has color options. Open the color options and confirm that the default palette and gradients are not available.
Insert and image block. In the block toolbar, select the duotone option. Confirm that the default duotones are not available.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -4,6 +4,9 @@
 	"settings": {
 		"appearanceTools": true,
 		"color": {
+			"defaultDuotone": false,
+			"defaultGradients": false,
+			"defaultPalette": false,
 			"palette": [
 				{
 					"color": "#FFFFFF",


juanfra (approved):
Thanks Carolina, looks good to me.

https://github.com/user-attachments/assets/c5533bcb-5096-4ff6-92be-2f02d186fa9b

Issue #305, opened 2024-09-13
Text styles


richtabor:
Any reason why we wouldn't leverage more block style variations? 

I noticed a number of uses of this for example: 

![CleanShot 2024-09-13 at 16 49 50](https://github.com/user-attachments/assets/111c0616-e3a6-46fe-8b48-be0bda5625bd)

(XL font, 1.2 line height)

If we can identify a few candidates, it would be nice to share styles across headings, paragraphs, perhaps even quotes.

carolinan:
Lets start with this one XL, 1.2 line height on the paragraph block, and test what the experience is like.

There are now **at least** 12 font sizes in the theme, and I think that replacing all 7 of the custom presets with style variations for the heading block would become too overwhelming. But to be clear, that is only my own subjective opinion.

richtabor:
Yea, we don't need all these expressed as text styles-only the combination of font, size, and line heights that are commonly used and expected.

PR #330, opened 2024-09-16
Add typography block style variations

carolinan:
**Description**
This is an experiment for https://github.com/WordPress/twentytwentyfive/issues/305

This does not replace the custom font sizes in theme.json, it is a complement that helps us remove hard coded font sizes in the markup.
This means that the size and line height can be adjusted in a global style variation, _if we want._

What was changed:
- A new block style variation for the paragraph block was added, with font size XL and line height 1.2. The font style was then applied to a section of a block pattern, replacing the size and line height in the markup.
- A new block style variation for the paragraph, heading and post-terms block was added, with font size small, uppercase, font-weight 700, and letter-spacing 1.4px.

**Screenshots**
This screenshot shows the "Lead" paragraph style variation, text with the old markup first, and the new variation second: There should be no visible difference:
p-xl

This video shows how the style is applied:
https://github.com/user-attachments/assets/af5eb091-3efa-412c-9c91-b8b5b1cbc8be

**Testing Instructions**

Place a paragraph, a post terms block and two heading blocks: One of the headings should use the H6 level.
Apply the different style variations, and compare the result in the editor and front.
The uppercase variation should match the style of the H6.


beafialho:
This looks good to me!

carolinan:
It's more a matter of if this is really something that we want to do. Is having two font size controls the best experience for the user.

juanfra (commented):
Personally, I am not sure about adding this one. I believe it could be confusing for users having two places competing for the same functionality, so close to each other and with the same name (XL). I believe, for many users, the difference between both of these won't be clear.

Are we going to be adding more of these?

beafialho:
I like that we're providing another styling option, but it can seem a bit incoherent to see two XL in two different places.

carolinan:
If a style variation adds more than just a size, for example size, letter-spacing, transformation, weight and line height combinations; then they are more helpful because they are like shortcuts, much like the "pill" or "rounded" styles.
If the user wants to adjust all of these typography settings on a block, it would take many clicks, with a variation it's just one click.

But Rich only pointed out this one combination, which is used in two patterns: this and one other.

What kind of names could this type of variation use?
"Ingress", "lead", "Intro", "Subheading" (for headings) comes to mind.
"Large lead" "Uppercase subheading" ?

carolinan:
Lede? Callout? Byline? Kicker? -Yes I asked Chat GPT for examples from the news world.

carolinan:
BTW these section style names still needs to be updated "Style 1" 😄

beafialho:
> If a style variation adds more than just a size, for example size, letter-spacing, transformation, weight and line height combinations; then they are more helpful because they are like shortcuts

Yes that's the best term, I like that we'd be providing shortcuts 👍

I find "Intro" or "Large Lead" more intuitive than the others.

carolinan:
This example uses the XL font size so maybe "Large lead" then. It depends on if we are adding more...

beafialho:
Another style I can think of we could add is this one:

Captura de ecrã 2024-09-17, às 09 43 36

 It has specific letter spacing and can work nicely in some occasions. It would be a nice shortcut to add.

It's used in some categories:
Captura de ecrã 2024-09-17, às 09 41 11
Captura de ecrã 2024-09-17, às 09 41 17

And H6:
Captura de ecrã 2024-09-17, às 09 41 28

styles/blocks/uppercase.json
```diff
@@ -0,0 +1,15 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"slug": "uppercase",
+	"title": "Uppercase",
```
carolinan, line 5: I also tried "Bold Uppercase" but it did not fit inside the button in the interface, it got truncated.

jasmussen:
Thanks for the exploration. Here's a quick GIF of the interface:

paragraph variations

My main comment here would be the user experience of using these variations for paragraphs. I think the instinct behind #305 (CC: @richtabor) is a good one: making it easy for someone to easily apply visual styles across text-blocks so as to easily achieve a coherence across content. 

The main problem with this is the variation interface itself. It has high prominence, being at the top of the inspector, which for a base atomic block such as the paragraph is pretty big. It's also not that valuable in most cases of authoring blog posts; all-caps and pill shaped notably are not going to make good reading experiences in those cases. Those styles are perfect for unique use cases such as subheadings or other section delimiters. 

Pending any additional thoughts by Rich, I'd lean towards _not_ adding these variations, at the very least not to the Paragraph block, since improvements to the variation UI are not immediately on the radar for the block editor. Would it be possible instead to have a category of utility patterns? Subheadings, things like that, which come with these all-caps and pill shape styles by default?

"Lead" is also very close to the XL default font size, even if it has a tighter line-height. I don't know that's sufficent to add this.

carolinan:
The problem with adding them as patterns is that all the styling is in the markup, so they can not be changed in the global style variations and section styles.

jasmussen:
Yep, that would be the trade-off, and not ideal. I think we should work to improve this, through type styles or otherwise, in the block editor itself. That's a longer term view.

carolinan:
Can we review the use of paragraphs with the pill style and uppercase again, and use the headings -**with the correct heading levels** as suggested and disregarded in past pull requests.

Using all uppercase for large sections of text reduces readability*, having these options for the paragraph may encourage unintended use.

https://accessibility.huit.harvard.edu/design-readability
https://uit.stanford.edu/accessibility/concepts/typography#allCaps

jasmussen:
> Using all uppercase for large sections of text reduces readability*, having these options for the paragraph may encourage unintended use.

Just to clarify, that's specifically one of the reasons why I'd suggest not making these into _block variations_. But that doesn't mean it's not useful as patterns, whether small utility patterns on their own, or as part of larger patterns, where these styles are used well.

carolinan:
After removing the paragraph styles, the theme is still left with the uppercase block style for the heading and post terms, is there value in keeping them? I did not finish implementing it since we were still discussing it, there are more places where using the CSS class could help remove the hard coded style from the markup.

richtabor:
I don't think we should have one-off text styles applicable to some content/text blocks, but not others. It creates an inconsistent experience. And it's not really solving the typography situation we have today, where some styles are hardcoded into paragraphs and headings (already).

PR #333, opened 2024-09-16
Adds Noon, Evening and Afternoon variations

beafialho:
**Description**

Adds combined variations remaining in #193. 

I had to change the hues and order of some color palettes for the section styles to work. I'll update the Figma with the new colors.

**Screenshots**

Evening:

Captura de ecrã 2024-09-16, às 16 48 12

Noon:

Captura de ecrã 2024-09-16, às 16 58 21

Afternoon:

Captura de ecrã 2024-09-16, às 16 59 15


carolinan:
The buttons especially in noon are still problematic, with some background colors both the default and outline buttons look like plain text.

carolinan:
From the screenshot, it looks like section 3 and 5 in afternoon is the same?

beafialho:
The button outlines in styles 2 and 4 don't look good in any of the variations. Is there something we can do about it?

beafialho:
I'll try to add a different color in Afternoon.

carolinan:
> The button outlines in styles 2 and 4 don't look good in any of the variations. Is there something we can do about it?

I think this will be the most difficult part about adding the variations.

In styles/blocks/section-2.json the button is set to contrast and base:

```

"elements": {
			"button": {
				"color": {
					"background": "var:preset|color|contrast",
					"text": "var:preset|color|base"
				}
			}
```

Then in noon (as an example), section-2 can be updated like this:

```
"variations": {
		"section-2": {
				"elements": {
					"button": {
						"color": {
							"background": "anything",
							"text": "anything"
						}
					}
...
```

carolinan:
(the outline button can not be changed the same way)

carolinan:
### Evening
I am not seeing the same results for section style 5 as in the screenshot above.
image
So I will change the text color for this section style in this variation.

carolinan:
> The button outlines in styles 2 and 4 don't look good in any of the variations. Is there something we can do about it?

Yes the colors can be changed. And I mean I can implement it and _I could make a decision about what colors to use_ but you might not like it 🥹 
So basically I need to know what it is _supposed_ to look like.

carolinan:
For **evening**, I believe section one is supposed to have buttons with black background:
image

And section two buttons should have a dark purple background:
image

Section 3 should have the white buttons that they do in this PR, but after that, I don't know because the styles are not in Figma.

carolinan:
Here is the list of color changes that were made: I would like to know if the color palette presets should be updated to match the changes in the global style variation.

### Noon
| Noon palette | Noon global style |
| Contrast `#5F5F5F`  | New contrast `#191919`  |
| Accent 2 `#914D0E`  | New accent 2 `#f5b684`  |

### Evening
| Evening palette | Evening global style |
| Contrast `#D1D1D1`  | New contrast `#f0f0f0`  |
| Accent 1 `#6F6720`  | New accent 1 `#786d0a`  |
| Accent 2 `#CFBEE4`  | New accent 2 `#442369`  |
| Accent 3 `#3F3272`  | New accent 3 `#d1d0ea`  |
| Primary `#A1A1A1`  | New Primary `#cbcbcb`  |

jasmussen:
Overall, looks good! Saw no issues with the styles other than what Carolina has mentioned. 

What @richtabor mentioned about style variations is still on my mind for this one: is it simply echoing the changes you are already proposing [here](https://github.com/WordPress/twentytwentyfive/pull/333#issuecomment-2367260954), @carolinan? 

Essentially what we want to avoid is having _style variations_ that have color or font palettes that are unique from the discrete color and font presets, these:

Screenshot 2024-09-23 at 12 38 25

If we can keep those in sync, we can move fast. Otherwise we should get the presets perfected first, and _then_ combine them into variations. What do you think? 

Note, I was AFK for a bit and catching up, so apologies if I'm missing a nuance, and thank you for the pings!

carolinan:
Yes, that was my intention, for the palette color and the global style variations to match. It is not clear to me why the palettes needed to change.

carolinan:
I believe it would be easier to manage these as separate pull requests.

carolinan:
In afternoon, section 5 is drastically different when comparing the pull request and Figma, so I have not touched it.

PR #334, opened 2024-09-16
Add: Sunrise style variation.

juanfra:
**Description**

Add the Sunrise global style variation.
Partial for https://github.com/WordPress/twentytwentyfive/issues/193

**Known limitations**

- Hover for outline buttons in sections.

**Screenshots**

Screenshot 2024-09-16 at 19-30-18 Style Variations – Stories®

**Testing Instructions**

1. Go to Appearance > Editor > Styles
2. At the top of the styles panel, select "Sunrise".
3. Save.
4. Compare the design with the design in Figma


carolinan:
Shall we merge this? It's the last one I think. (Though all of them will need some iterations).

juanfra:
Sounds good to me. We can ask Bea to make a pass on all of them once she's back.

However, please let me know if you find something that needs improvement and can be addressed now. 

Thanks, Carolina!

carolinan:
The hover styles for the default (filled) buttons when they are placed in a group with a section style is not implemented.

There are two color contrast issues: the link colors in section style 1 and 2.
I know that the link color information is missing from Figma. Until the _final final_  design review can happen, I updated the other variations that had the same problem, and used the text color for the link color.

carolinan:
I also know that we have not discussed the naming of the files yet.
I assumed that @beafialho would like the combined style variations to have the same order as the color palettes, so I named the other files with `01-name` and so on.

juanfra:
Thanks Carolina, I worked on the issues you've raised.

carolinan:
For a follow up: I _think_ the search block button should have the same radius as the other buttons.

PR #335, opened 2024-09-16
update: added i18n to index & home template headings

### LINKED ISSUE #323: Task: Translatable text in the HTML files must be turned into patterns
**Description**
All HTML files need to be reviewed to locate and fix any remaining text strings.
**All** the text in the theme needs to be translation ready, and text in HTML files can't be translated.
This includes texts inside attributes, such as placeholders and aria labels, and any other default texts that are added when a block is inserted in the editor.

If you find that a pattern is needed, **please check carefully if there is an existing pattern** that can be inserted in the template, before creating a new.

up1512001:
fixes https://github.com/WordPress/twentytwentyfive/issues/323

**Description**
- created a hidden pattern for the heading of the home & index template.
- added i18n to the `Blog` text.
- replace the core/heading block of the home & index template with the hidden pattern.

**Screenshots**

https://github.com/user-attachments/assets/85a3d78f-f300-490a-acb7-843a5811279a

### DIFF
--- a/patterns/hidden-blog-heading.php
+++ b/patterns/hidden-blog-heading.php
@@ -1,7 +1,7 @@
 <?php
 /**
- * Title: Hidden index home page heading
- * Slug: twentytwentyfive/hidden-index-home-page-heading
+ * Title: Hidden Blog Heading
+ * Slug: twentytwentyfive/hidden-blog-heading
  * Description: Hidden heading for the home page and index template.
  * Inserter: no
  *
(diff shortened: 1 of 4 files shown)


carolinan:
The PHP file needs to end with a single empty line.
Can we use a shorter name and slug? Why not just use hidden-blog-heading ?

This makes both headings left aligned, but from the recording, this does not seem to cause any issues.

up1512001:
@carolinan updated file name & slug as per your suggestion.

up1512001:
> This makes both headings left aligned, but from the recording, this does not seem to cause any issues.

@carolinan Yes, aligning the heading to the left for `index.html` is not causing any issues.

carolinan (approved):
Thank you for working on this :)

PR #336, opened 2024-09-16
Add: theme utils to escape patterns.

juanfra:
**Description**

Adding theme utils to escape patterns via command. It uses a variation of the `theme-utils` from community themes, with some hardcoded pieces to work with TT5. This iteration only includes the command to escape patterns and not the json schema validator.

To try it out, first, you need to install it.

```
nvm use && npm install
```

Once installed, you can try it out using the following command. It only works on the files that are modified in your local git.

```
npm run patterns:escape
```

It'll be a good tool to speed up escaping the different patterns we have already merged. We should have probably looked into this before, but it's better now than never.

**Screenshots**

https://github.com/user-attachments/assets/a9534047-ad67-4ef5-9f8c-912d791f08fd

**Testing Instructions**

See description.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,8 @@
 			"license": "GPL-2.0-or-later",
 			"devDependencies": {
 				"@wordpress/stylelint-config": "^22.5.0",
+				"fs-extra": "^11.1.0",
+				"parse5-html-rewriting-stream": "^7.0.0",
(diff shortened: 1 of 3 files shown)


juanfra:
Closing in favor of https://github.com/WordPress/twentytwentyfive/pull/358

PR #340, opened 2024-09-17
Fine tuning adjustments to default header and footer

beafialho:
**Description**

Fine-tuning adjustments to the default header and footer:

- Reduced top and bottom padding of the header
- Set tagline to `M` instead of `S` font size in theme.json
- Adjustments to footer for better scaling across devices

### DIFF
--- a/patterns/footer.php
+++ b/patterns/footer.php
@@ -12,66 +12,56 @@
  */
 
 ?>
-<!-- wp:group {"metadata":{"categories":["footer"],"patternName":"twentytwentyfive/footer","name":"Footer"},"style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|50","left":"var:preset|spacing|50","right":"var:preset|spacing|50"},"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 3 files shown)


carolinan:
Footer before:

https://github.com/user-attachments/assets/708dc3fd-813e-47dc-b657-3239ae089f8c

After:

https://github.com/user-attachments/assets/286437da-6cab-49f8-b50e-ee0861925266

Is the site title and tagline  supposed to be that narrow and wrap like that on the desktop width?

Front, after:
footer

carolinan (changes requested):
There is a block validation error in this header.
There is a group that is incomplete or not closed.

beafialho:
> Is the site title and tagline supposed to be that narrow and wrap like that on the desktop width?

No 😕

**Here's what I'm seeing in the editor:**

Captura de ecrã 2024-09-17, às 15 29 27

**And on the front end:**

Captura de ecrã 2024-09-17, às 15 29 56

beafialho:
> There is a block validation error in this header. There is a group that is incomplete or not closed.

I tried to change only what was necessary so I wouldn't touch the code, but I guess I missed something. I submitted another change to the footer to hopefully fix that.

carolinan:
The problem with the broken group block is in the header, not the footer. I'll try to help another time when I am more alert :)

beafialho:
> The problem with the broken group block is in the header, not the footer. I'll try to help another time when I am more alert :)

Oops 🙈 this is what happens when the designer touches code.

The only thing I wanted to change in the header was to set the top and bottom spacing of the row to `spacing-30` instead of `spacing-40`.

Thank you for the help, no urgency!

juanfra (approved):
Thanks, Bea!

I've fixed the block validation issue and worked on the indent of the footer.

Looks good to me

https://github.com/user-attachments/assets/956a2f72-049d-4435-a4f7-9833cf80f40e

Issue #302, opened 2024-09-13
Pattern review


richtabor:
### A review of current patterns: 

- [x] This needs better mobile: 

![CleanShot 2024-09-13 at 16 34 31](https://github.com/user-attachments/assets/8ffe488f-2a83-4e2b-9c42-5c64f3a3a6f7)

- [x]  Let's make this pattern's grid minimum value `26` so that it can render as designed a bit longer, before switching to stacked layout on mobile: 

![CleanShot 2024-09-13 at 16 35 32](https://github.com/user-attachments/assets/4d1b4d49-901f-43a8-9511-633988ab1113)

- [x] Instead of one-side vertical padding, let's use spacer blocks. It makes the selected block look proper when it's contained properly and spacer blocks are easier to manipulate in scenarios like this below. This padding also needs the horizontal padding removed (https://github.com/WordPress/twentytwentyfive/issues/300). 

![CleanShot 2024-09-13 at 16 37 48](https://github.com/user-attachments/assets/bb8b2d17-cba4-4294-9f4b-d9c998e178a8)

- [ ] If we cannot resolve the slow scaling issue with this pattern, we should edit it to not use 100vh: 

![CleanShot 2024-09-13 at 16 39 32](https://github.com/user-attachments/assets/1e7a9e32-305d-4e37-89ce-125d0b8a7cc9)

- [x] This is another scenario where a spacer block is better suited. Block gap is only affecting one space (between heading and grid). It's also buried in the Inspector:

![CleanShot 2024-09-13 at 16 40 49](https://github.com/user-attachments/assets/810f00a6-12cc-4cb1-b71f-0f3e204baf44)

- [x] Remove padding bottom here, use a spacer instead: 

![CleanShot 2024-09-13 at 16 47 01](https://github.com/user-attachments/assets/a1c47fae-936e-45ab-a869-5b89caaa8f96)

- [x] I suggest removing this one, it's not working well as-is: 
![CleanShot 2024-09-13 at 16 50 44](https://github.com/user-attachments/assets/0cfbca8d-9a68-4754-888c-e4a6c95fa3ac)

carolinan:
Please, kindly explain what "it's not working well as-is:" means. Otherwise this is not possible to address.

carolinan:
Here is a recording of the scaling issue in the "Link in bio with tight margins" pattern:

https://github.com/user-attachments/assets/4ebee8ec-65dc-4b5d-8aaf-478f5bb0b375

[...]

PR #343, opened 2024-09-17
Pattern review: CTA Pricing, clients and banner with description patterns.

juanfra:
**Description**

Partial to https://github.com/WordPress/twentytwentyfive/issues/302

- CTA Pricing pattern: Removed padding and used spacer instead.
- Clients: Removed padding and used spacer instead. Switched heading from H3 to H2 to match design.
- Banner with description: Set minimum column width to 26.

**Screenshots**

N/A

**Testing Instructions**

CTA pricing pattern

1. Create a page.
2. Use the CTA pricing pattern.
3. Confirm the padding is no longer there for the stack and we're using a spacer instead.

Clients pattern

1. Create a page
2. Use the clients pattern.
3. Confirm that we're no longer using padding and we are now using a spacer.
4. As a bonus, the H3 is now an H2, to respect the Figma design.

Banner with description

1. Create a page.
2. Use the "Banner with description and images grid"
3. Confirm that the grid minimum col width is set to 26.


patterns/cta-pricing.php
```diff
@@ -26,6 +26,10 @@
 	</div>
 	<!-- /wp:group -->
 
+	<!-- wp:spacer {"height":"var:preset|spacing|50"} -->
+	<div style="height:var(--wp--preset--spacing--50)" aria-hidden="true" class="wp-block-spacer"></div>
+	<!-- /wp:spacer -->
```
carolinan, line 31: Because the spacer block has the default top margin of a nested block, spacing preset 50 on the block is larger than the previous bottom padding.

Left is the current, right is the proposed change in the PR:

image

juanfra: Thanks, Carolina. I added the changes accounting for that margin.

Issue #344, opened 2024-09-17
The two sidebar patterns are identical: remove one and keep one


carolinan:
The two sidebar patterns twentytwentyfive/hidden-sidebar and twentytwentyfive/right-aligned-sidebar are identical except for the indentation of the code and the capitalization of the heading above the query.

Lets remove one pattern and use the remaining pattern for both sidebars.

ikamal7:
Hi @carolinan 
I would like to address this issue by working on it. The two sidebar patterns, `twentytwentyfive/hidden-sidebar` and `twentytwentyfive/right-aligned-sidebar`, are indeed identical except for the code indentation and the capitalization of the heading above the query.

To streamline, I propose removing the `right-aligned-sidebar` pattern and using the remaining pattern for both sidebars. Specifically, we can solve the issue by removing the pattern `<!-- wp:pattern {"slug":"twentytwentyfive/right-aligned-sidebar"} /-->` from `right-aligned-sidebar.html` and including the generic sidebar pattern instead `<!-- wp:pattern {"slug":"twentytwentyfive/hidden-sidebar"} /-->`.

Let me know if there are any additional steps or considerations to be aware of.

carolinan:
That sounds accurate 😊

ikamal7:
Should I remove the`twentytwentyfive/patterns/vertical-header-right-aligned-sidebar.php` file as well?

PR #347, opened 2024-09-17
remove identical sidebar pattern

ikamal7:
**Description:**

This PR resolves issue #344 by removing the duplicate `twentytwentyfive/right-aligned-sidebar` pattern. The `twentytwentyfive/hidden-sidebar` pattern is retained and used for both sidebars. The pattern `<!-- wp:pattern {"slug":"twentytwentyfive/right-aligned-sidebar"} /-->` has been removed from `right-aligned-sidebar.html`.


carolinan (approved):

juanfra:
Why are we keeping two files with the same purpose?

`parts/sidebar.html` and `parts/right-aligned-sidebar.html` are the same. We should keep only one, I'd keep only `parts/sidebar.html` and replace the appearances of

```
<!-- wp:template-part {"slug":"right-aligned-sidebar"} /-->
```

with

```
<!-- wp:template-part {"slug":"sidebar"} /-->
```

PR #349, opened 2024-09-17
Update text-wrap style. Closes #346

### LINKED ISSUE #346: Enhancement to reduce widows and orphans in the theme
Sometimes text can have single lines of text (widows and orphans) from a paragraph that dangle at either the beginning or end of a block of text, which blocks readability. Unless [this Gutenberg issue](https://github.com/WordPress/twentytwentyfive/issues/25#issuecomment-2315517591) gets implemented, one solution could be adding the code below (or something similar that achieves the same) in `style.css` to prevent that from happening.

image

```
/* Progresive enhancement to reduce widows and orphans.
/* <https://github.com/WordPress/gutenberg/issues/55190>
*/
h1, h2, h3, h4, h5, h6, blockquote, caption, figcaption {
		text-wrap: balance; /* Safari fallback */
		text-wrap: pretty;
}
p {
    text-wrap: pretty;
}

```

bobmatyas:
**Description**

Per #346, this adds the `text-wrap` style to deal with widows and orphans.

**Screenshots**

Before Example:

before-text-wrap

After Example:

after-text-wrap

**Testing Instructions**


style.css
```diff
@@ -28,3 +28,16 @@ a {
 	outline-width: 2px;
 	outline-style: solid;
 }
+
+/*
+ * Progresive enhancement to reduce widows and orphans
+ * https://github.com/WordPress/gutenberg/issues/55190
+ */
+h1, h2, h3, h4, h5, h6, blockquote, caption, figcaption {
+	text-wrap: balance; /* Safari fallback */
```
ryelle: `balance` can be an issue for CJK languages (and possibly others) ([see this core ticket](https://core.trac.wordpress.org/ticket/60892)). If it's just a fallback, I would suggest removing it and letting the fallback be default wrapping.

bobmatyas: @ryelle Thanks for letting me know. I followed the suggested approach on the GH issue and wasn't aware of the core ticket. I have removed `balance`.

carolinan (changes requested):
Hi, please remove changes that are unrelated to the text-wrap.

bobmatyas:
@carolinan I think I was able to remove the extraneous commits. 🤞

style.css
```diff
@@ -28,3 +28,11 @@ a {
...
 	outline-style: solid;
 }
+
+/*
+ * Progresive enhancement to reduce widows and orphans
+ * https://github.com/WordPress/gutenberg/issues/55190
+ */
+h1, h2, h3, h4, h5, h6, blockquote, caption, figcaption, p {
+	text-wrap: pretty;
+}
```
carolinan:
```suggestion
}

```
There needs to be a single blank line at the end of the document.

carolinan (approved):

carolinan:
This style is only enqueued on the front, and I think this is the right decision.
Without extensive testing, we won't know if there would be any negative impacts on the interface if the style would be added to the editor. Some parts of the editor already use text-wrap on these elements.

PR #350, opened 2024-09-17
Removes 'more posts' from home and index templates

### LINKED ISSUE #326: Home: The "more posts" lists the same posts as the main query.
**Description**

On the default home template, the main query is set to inherit from the template (and this is correct) which means that
the number of posts shown is controlled by the setting in the admin area: Settings > Reading Settings > Blog pages show at most.
This means that the second query under the heading "More posts" shows the title and date of the same posts.
(Once you navigate to the second page of the result, this is a non issue, it is only a problem on the first page).

home-more-posts

There is no way to offset this query to same number as in the reading settings of the first query.
By changing the sorting order and use for example oldest to newest, the posts may be different, but it would also promote old content, which may be undesirable.

bobbylee-noble:
**Description**

The Home and Index templates on twentytwentyfive have a section(pattern) that lists 'More Posts'. This is unnecessary and shows the first posts on all pages when using pagination. One the first page this is also redundant. 
#326

### DIFF
--- a/templates/home.html
+++ b/templates/home.html
@@ -4,7 +4,6 @@
 <main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
 	<!-- wp:pattern {"slug":"twentytwentyfive/hidden-blog-heading"} /-->
 	<!-- wp:pattern {"slug":"twentytwentyfive/posts-personal-blog"} /-->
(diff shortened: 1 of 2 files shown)


carolinan (approved):
Thank you for working on this!

bobbylee-noble:
@carolinan Thanks! this is exciting! Should I delete this branch, what is the process here?

PR #351, opened 2024-09-17
removing query no results patterns

### LINKED ISSUE #319: Remove the "no results" text patterns
**Description**
The theme has two patterns for the "No results" text, that are already translation ready: One with default alignment and one that is centered.

hidden-no-results-centered.php: twentytwentyfive/hidden-no-results-centered
hidden-no-results.php: twentytwentyfive/hidden-no-results

But these patterns are only used in two query loops.
There are 22 (!) query loops where the patterns are not used.

Even without testing, I assume that including and loading the pattern 20+ times would not be performant compared to just adding the correct translation strings to each query.

As long as the exact same text string is used for every no results block, it should only count as one translatable string when the contributors translate the theme, even though it is added over 20 times. -Please correct me if that is wrong.

mathew-crisp:
**Description**

There are two "no results found" patterns, "hidden-no-results-centered.php" and "hidden-no-results.php", which are used only once each. Request is to replace the patterns where used with the translatable text instead to increase performance.
#319

**Testing Instructions**

1. Activate the 2025 theme.
2: Two patterns are updated, 
     "Photo blog posts" where the "No posts were found." text should be centered
     "List of posts, 1 column" where the "No posts were found." text should be left-aligned
3: For both of the above, ensure the "No posts were found." appears when appropriate, and does not show when posts are found.


carolinan (approved):
The patterns were added with a good intent, to not repeat code.
But since using the patterns within all the query loops can potentially affect performance,
removing them seems to be right choice.

Thank you for working on this problem. 
If you connect your GitHub and WordPress org accounts, please add a comment, then the "props" can be updated.

mathew-crisp:
Wordpress.org account should be linked by now, mathew-crisp

juanfra:
Thank you, @mathew-crisp - to link your GitHub account with WP.org, you should follow the steps described [here](https://make.wordpress.org/core/handbook/tutorials/linking-your-github-and-w-org-profiles/).

Issue #160, opened 2024-08-29
Translation issue: All the text needs to be translation ready


codersantosh:
**Description**
I've noticed a few strings that need translation on patterns and templates. Are these currently under review for finalization, and will the pattern designer push the final version? Also, how do you plan to translate the text within .html files?

carolinan:
Translations will be finalized later before the 6.7 string freeze.
This is because there will be several changes to the templates and pattern files while the theme is developed.


The problems with adding the translations early is that
- the PHP code creates block validation error if it is not copied correctly
- the PHP is removed every time a change is made in the editor, so re-adding the translation functions over and over is a time-sink that should be avoided.


The theme does not add any extra functionality for translations. The text needs to be added in the PHP files.

codersantosh:
Thank you, @carolinan, for answering all my questions. Best of luck to you and the team on the project.

carolinan:
I think it is fine to keep this open as this will be a task later.

carolinan:
### What is internationalization?
Internationalization is the process of developing your theme, so it can easily be translated into other languages. 

https://developer.wordpress.org/apis/internationalization/
https://developer.wordpress.org/apis/internationalization/internationalization-guidelines/

carolinan:
Pattern titles and descriptions do not need to be wrapped in the PHP translation functions. This is handled elsewhere in WordPress:
```
/**
 * Title: Services - Three Columns
 * Description: Displays three columns with images and text for services.
 *
```

These type of texts need to be made translatable:
```
	<!-- wp:heading {"align":"wide"} -->
	<h2 class="wp-block-heading alignwide">Our services</h2>
	<!-- /wp:heading -->
```

`alt="Image for service"`

In this case, the placeholder should be removed because it is not used.  And because there are multiple of these no results blocks in the theme, they should all use identical translatable strings for the paragraph.
```
	<!-- wp:query-no-results -->
		<!-- wp:paragraph {"placeholder":"Add text or blocks that will display when a query returns no results."} -->
		<p>No posts were found.</p>
		<!-- /wp:paragraph -->
	<!-- /wp:query-no-results -->
```

In the search block, both the label, placeholder, and button text needs to be translation ready:

[...]

PR #355, opened 2024-09-17
Update/translation strings wcus 160

bobmatyas:
**Description**

Adds translations to strings in `/patterns/`:

- "Posted by"
- "Published on"
- "Tagged"
- "Categories"

Part of https://github.com/WordPress/twentytwentyfive/issues/160

### DIFF
--- a/patterns/photo-blog-single-post-template.php
+++ b/patterns/photo-blog-single-post-template.php
@@ -34,14 +34,14 @@
 						<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|10"}},"layout":{"type":"constrained"}} -->
 						<div class="wp-block-group">
 							<!-- wp:paragraph {"fontSize":"small"} -->
-							<p class="has-small-font-size"><?php esc_html__( 'Published on' ); ?></p>
+							<p class="has-small-font-size"><?php esc_html__( 'Published on', 'twentytwentyfive' ); ?></p>
 							<!-- /wp:paragraph -->
 							<!-- wp:post-date {"style":{"elements":{"link":{"color":{"text":"var:preset|color|contrast"}}}},"textColor":"contrast","fontSize":"small"} /--></div>
 						<!-- /wp:group -->
 						<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|10"}},"layout":{"type":"constrained"}} -->
 						<div class="wp-block-group">
 							<!-- wp:paragraph {"fontSize":"small"} -->
-							<p class="has-small-font-size"><?php esc_html__( 'Posted by' ); ?></p>
+							<p class="has-small-font-size"><?php esc_html__( 'Posted by', 'twentytwentyfive' ); ?></p>
 							<!-- /wp:paragraph -->
 							<!-- wp:post-author-name {"isLink":true,"fontSize":"small"} /-->
 						</div>
@@ -61,7 +61,7 @@
 						<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
 						<div class="wp-block-group">
 							<!-- wp:paragraph {"fontSize":"small"} -->
-							<p class="has-small-font-size">Tagged:</p>
+							<p class="has-small-font-size"><?php esc_html__( 'Tagged', 'twentytwentyfive' ); ?>:</p>
 							<!-- /wp:paragraph -->
 							<!-- wp:post-terms {"term":"post_tag","style":{"typography":{"fontStyle":"normal","fontWeight":"300"}}} /-->
 						</div>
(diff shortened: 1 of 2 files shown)


patterns/photo-blog-single-post-template.php
```diff
@@ -34,14 +34,14 @@
 						<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|10"}},"layout":{"type":"constrained"}} -->
 						<div class="wp-block-group">
 							<!-- wp:paragraph {"fontSize":"small"} -->
-							<p class="has-small-font-size">Published on</p>
+							<p class="has-small-font-size"><?php esc_html__( 'Published on', 'twentytwentyfive' ); ?></p>
```
juanfra: [esc_html__()](https://developer.wordpress.org/reference/functions/esc_html__/) only retrieves the translation but doesn't print it. We should use either `echo esc_html__()` or [esc_html_e()](https://developer.wordpress.org/reference/functions/esc_html_e/)

juanfra (commented):
Thank you for working on this. I left a comment for you to check, as with the functions this PR is using, the text won't be printed.

patterns/photo-blog-single-post-template.php
bobmatyas: @juanfra There was some confusion on the functions at WCUS Contributor Day. I have updated them and re-requested review.

patterns/photo-blog-single-post-template.php
```diff
@@ -53,15 +53,15 @@
 						<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
 						<div class="wp-block-group">
 							<!-- wp:paragraph {"fontSize":"small"} -->
-							<p class="has-small-font-size">Categories:</p>
+							<p class="has-small-font-size"><?php esc_html_e( 'Categories', 'twentytwentyfive' ); ?>:</p>
```
juanfra: Let's keep the `:` to match the design.

patterns/photo-blog-single-post-template.php
```diff
@@ -53,15 +53,15 @@
...
 							<!-- /wp:paragraph -->
 							<!-- wp:post-terms {"term":"category","style":{"typography":{"fontStyle":"normal","fontWeight":"300"}}} /-->
 						</div>
 						<!-- /wp:group -->
 						<!-- wp:group {"style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
 						<div class="wp-block-group">
 							<!-- wp:paragraph {"fontSize":"small"} -->
-							<p class="has-small-font-size">Tagged:</p>
+							<p class="has-small-font-size"><?php esc_html_e( 'Tagged', 'twentytwentyfive' ); ?>:</p>
```
juanfra: Let's keep the `:` to match the design.

juanfra (commented):
Thank you for adjusting the translation functions. I left a few comments to keep some texts as they were. Thanks!

patterns/photo-blog-single-post-template.php
bobmatyas: @juanfra I'm unsure about this one. It looks like the `:` is added after the `Catgories` and `Tagged` strings in the PR:

https://github.com/WordPress/twentytwentyfive/pull/355/files#diff-d8a2a54e1164459ea5b389207d5c9493b8f09d3473f5aeb4f98563a72376a9bf

Happy to fix if needed, but it seems like the change is already present.

juanfra: Thanks, I meant inside the translation. We can leave it as it is right now, though.

juanfra:
```suggestion
							<p class="has-small-font-size"><?php esc_html_e( 'Categories:', 'twentytwentyfive' ); ?></p>
```

patterns/photo-blog-single-post-template.php
juanfra:
```suggestion
							<p class="has-small-font-size"><?php esc_html_e( 'Tagged:', 'twentytwentyfive' ); ?></p>
```

juanfra (approved):
Thanks so much!

patterns/photo-blog-single-post-template.php
carolinan: For the translators there is a difference between "Categories" and "Categories:"
It is not 100% obvious that "Categories" in this instance is followed by a list of categories.

Normally I would suggest using the prefix setting on the post terms block to add the word "Categories:"
But because "Categories:" is on a separate line, there would need to be a line break in the attribute and the translatable string.

So my suggestion would be to use https://developer.wordpress.org/reference/functions/esc_html_x/:

`<p class="has-small-font-size"><?php echo esc_html_x( 'Categories:', 'Prefix before one or more categories. Categories: category name' , 'twentytwentyfive' ); ?>:</p>`

In general block themes are more difficult to translate, because like in this example, some languages will have different words or forms depending on if there is one category, two, or more categories.
In PHP based themes we can work around that with the sprintf and the plural forms.

PR #357, opened 2024-09-17
Make string in search template translatable. Re: #160

bobmatyas:
**Description**

The only template file that needed to be updated was `search.html`. I added the translations there.

### DIFF
--- a/patterns/hidden-search.php
+++ b/patterns/hidden-search.php
@@ -11,4 +11,4 @@
 
 ?>
 
-<!-- wp:search {"label":"<?php esc_html_e( 'Search', 'twentytwentyfive' ); ?>","showLabel":false,"buttonText":<?php echo esc_attr_x( 'Search', 'button label', 'twentytwentyfive' ); ?>","style":{"border":{"radius":"50px"}},"fontSize":"medium"} /-->
+<!-- wp:search {"label":"<?php esc_html_e( 'Search', 'twentytwentyfive' ); ?>","showLabel":false,"buttonText":"<?php echo esc_attr_x( 'Search', 'button label', 'twentytwentyfive' ); ?>","style":{"border":{"radius":"50px"}},"fontSize":"medium","borderColor":"opacity-20"} /-->
--- a/patterns/hidden-search.php
+++ b/patterns/hidden-search.php
@@ -1,9 +1,8 @@
 <?php
 /**
  * Title: Search
- * Slug: twentytwentyfive/search
- * Categories: search
- * Description: Translatable search form
+ * Slug: twentytwentyfive/hidden-search
+ * Inserter: no
  *
  * @package WordPress
  * @subpackage Twenty_Twenty_Five
--- a/patterns/search.php
+++ b/patterns/search.php
@@ -1,0 +1,15 @@
+<?php
+/**
+ * Title: Search
+ * Slug: twentytwentyfive/search
+ * Categories: search
+ * Description: Translatable search form
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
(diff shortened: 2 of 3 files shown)


luminuu:
@bobmatyas Just checked on this, you've added the translation function in a HTML file, which is not going to work. You'd need to move this to a pattern as PHP file and then call that pattern inside the `search.html`.

bobmatyas:
@luminuu Shoot, I misunderstood how that what was done. I'll make a PHP pattern for it.

patterns/search.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * Title: Search
+ * Slug: twentytwentyfive/search
+ * Categories: search
+ * Description: Translatable search form
```
juanfra: This one needs to be hidden. We need to rename the file to `hidden-search` and change the docblocks. For hidden patterns we can avoid the description.

```suggestion
 * Title: Search
 * Slug: twentytwentyfive/hidden-search
 * Inserter: no
```

templates/search.html
```diff
@@ -3,7 +3,7 @@
 <!-- wp:group {"tagName":"main","style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
 <main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
 	<!-- wp:query-title {"type":"search"} /-->
-	<!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search","style":{"border":{"radius":"50px"}},"fontSize":"medium"} /-->
+	<!-- wp:pattern {"slug":"twentytwentyfive/search"} /-->
```
juanfra: We'll need to update this one once we change the filename.

```suggestion
	<!-- wp:pattern {"slug":"twentytwentyfive/hidden-search"} /-->
```

juanfra (changes requested):
Thank you for working on this @bobmatyas - I left a few comments we need to address before merging.

patterns/search.php
bobmatyas: @juanfra this change has been made. I didn't know about the hidden pattern syntax. Thanks for explaining :)

templates/search.html
bobmatyas: @juanfra this change has been made. I didn't know about the hidden pattern syntax. Thanks for explaining :)

patterns/hidden-search.php
```diff
@@ -0,0 +1,14 @@
...
+ * Inserter: no
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+
+<!-- wp:search {"label":"<?php esc_html_e( 'Search', 'twentytwentyfive' ); ?>","showLabel":false,"buttonText":<?php echo esc_attr_x( 'Search', 'button label', 'twentytwentyfive' ); ?>","style":{"border":{"radius":"50px"}},"fontSize":"medium"} /-->
```
juanfra: There's a missing quote at the start of the `ButtonText` value.

juanfra (commented):
Thank you @bobmatyas - There's one last thing I found and this one should be good to merge.

patterns/hidden-search.php
juanfra: Let's fix the `buttonText`. And we could also set the `borderColor` to match the design, now that we're touching these files. Thanks, Bob!

```suggestion
<!-- wp:search {"label":"<?php esc_html_e( 'Search', 'twentytwentyfive' ); ?>","showLabel":false,"buttonText":"<?php echo esc_attr_x( 'Search', 'button label', 'twentytwentyfive' ); ?>","style":{"border":{"radius":"50px"}},"fontSize":"medium","borderColor":"opacity-20"} /-->
```

bobmatyas: Fixed below:

bobmatyas:
@juanfra Changes made in https://github.com/WordPress/twentytwentyfive/pull/357/commits/0c12404c8190ef69352acb2403bf3ba9ee668517

juanfra (approved):
Thank you @bobmatyas!

PR #359, opened 2024-09-17
Add theme preview with WordPress Playground

vcanales:
**Description**

I've recently updated the GitHub action that generates links to a Playground instance in order to preview the changes on a PR to [support single-theme repos](https://github.com/vcanales/action-wp-playground-pr-preview/pull/31), so that we may use it in repos like this one.

It still requires a lot polish, but after merging, we should be seeing messages such as the ones displayed in this PR from a fork I used for testing:

https://github.com/vcanales/twentytwentyfive/pull/2

### DIFF
--- a/patterns/hidden-404.php
+++ b/patterns/hidden-404.php
@@ -44,4 +44,3 @@
 	<!-- /wp:columns -->
 </div>
 <!-- /wp:group -->
-<?php echo 'Test Change'; ?>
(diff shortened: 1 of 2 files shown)


juanfra (commented):
Thanks Vicente! Works well

https://github.com/user-attachments/assets/ba231195-574d-478c-bc98-9702cf78d596

patterns/hidden-404.php
```diff
@@ -44,3 +44,4 @@
 	<!-- /wp:columns -->
 </div>
 <!-- /wp:group -->
+<?php echo 'Test Change'; ?>
```
juanfra:
```suggestion
```

PR #372, opened 2024-09-19
Enhancement: Make footers translatable.

juanfra:
**Description**

I am adding i18n functions for footers.

I'm still doubtful we should include `Twenty Twenty-Five` as a hardcoded string in some footers. I know this discussion happened before. I believe that now that we have the copyright block binding, that's a good candidate. Else, re-consider the site title.

**Screenshots**

N/A

**Testing Instructions**

1. Open the site editor
2. Try every footer
3. Check that the strings are correct.


patterns/footer-centered.php
```diff
@@ -23,7 +24,17 @@
...
-	<p class="has-text-align-center has-small-font-size">Designed with <a href="https://wordpress.org">WordPress</a></p>
+	<p class="has-text-align-center has-small-font-size">
+		<?php
+		/* Translators: WordPress link. */
+		$wordpress_link = '<a href="' . esc_url( __( 'https://wordpress.org', 'twentytwentyfive' ) ) . '" rel="nofollow">WordPress</a>';
+		printf(
+			/* Translators: Designed with WordPress */
+			esc_html__( 'Designed with %1$s', 'twentytwentyfive' ),
+			$wordpress_link // phpcs:ignore
+		);
```
carolinan, line 32: What is the purpose of using a variable when it is only used once?

juanfra: I'd say the purpose is readability. TT4 is using it like this. We can avoid the variable if you prefer it that way.

carolinan: By avoiding it, we could also avoid the `phpcs:ignore`.

juanfra: That's true, although there's nothing wrong with using `phpcs:ignore` per se. I'll apply the chances. I consider that this way is more readable, but that may be subjective.

Thanks.

carolinan:
I agree that the theme name should not be in the footer. It is common in commercial themes, but in this case, I don't think WordPress needs to strengthen the "Twenty Twenty-Five" name, and the WordPress credit is already in the footer. But I'm not a marketer.

carolinan:
This can be merged, if we want to handle the theme name text in a separate issue.

carolinan (approved):

juanfra:
Thanks, @carolinan. I believe we can keep that discussion on an issue to hear different perspectives and merge this one.

PR #374, opened 2024-09-19
Fix: Remove top and bottom margin 0 on full width patterns

### LINKED ISSUE #328: Remove top and bottom margin 0 on full width patterns
During development of the full width patterns, top and bottom margin was set to 0 because when the pattern has a background color, the sections are not "seamless": When the default global spacing between blocks is applied, there is a gap between the two colored sections:
gap between two patterns, where the body background is seen

But removing margins had the side effect that the block is too close to other type of content:
no-top-margin

There is no way to only remove the margin when two full width blocks follow eachother, without using custom CSS:
the better way would be for the user wrap to the two patterns inside another group and set the block spacing to 0.

I think the best would be to remove the margin setting from the patterns, and then on the full page patterns, where these full width patterns are combined into one cohesive layout, wrap them inside an extra group with block spacing set to 0.

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/328

This PR removes the margin top/bottom set to zero for full-width patterns.
It also updates the landing pages to be contained by a group with a block gap of zero.

Bonus:

- Add description to testimonials patterns.
- Indent several patterns.

**Screenshots**

N/A

**Testing Instructions**

1. Create a page.
2. Add the following patterns
   - Banner with book description
   - Banner with description and images grid
   - Clients section
   - Centered link and social links
   - Contact: Info and locations
   - Cover with big heading
   - Call to action with book links
   - Call to action with locations
   - Pricing two columns
   - Events – List with Call to Action
   - Events three columns with event images and titles
   - Grid with videos
   - Heading and paragraph with image on the right
   - Hero Centered Heading
   - Hero podcast
   - Link in bio with tight margins
   - Newsletter Sign Up
   - Poster-like section
   - Services - Subscriber only section
   - Services - Three-column boxed pricing table
   - Services - Three Columns
   - Review with large image on right
   - Three column layout with 6 testimonials
   - Two columns with avatar
   - FAQs
   - Two Headings and a List
3. Confirm that these patterns no longer have a margin top/bottom set to zero.


carolinan (approved):
I have followed the testing instructions and tested all the patterns in the list and the PR works as intended, thank you.

PR #377, opened 2024-09-20
Fix: Fine tune search block, and use it in different search results templates.

### LINKED ISSUE #184: Fine tune all search blocks
**Description**
Currently the patterns that use search forms are using 3 different styles:

`<!-- wp:search {"label":"Search","showLabel":false,"widthUnit":"%","buttonText":"Search","style":{"border":{"radius":"100px"}},"fontSize":"small","borderColor":"opacity-20"} /-->`

`<!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search","style":{"border":{"radius":"50px"}},"fontSize":"medium","borderColor":"opacity-20"} /-->`

`<!-- wp:search {"label":"Search","showLabel":false,"buttonText":"Search","style":{"border":{"radius":"50px"}},"fontSize":"medium"} /-->`

Lets try to find a way to remove the border radius from the block markup, and make sure that the border radius and border color matches. This is needed because when we start adding the style variations, they need to be able to override the styles of the inputs. See for example: https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=77-650&m=dev

Search form blocks that do not have a visible label, still needs the "label" attribute to contain the word "Search" and this should also be translation ready.

Two of the blocks uses font size medium and two uses font size small. The default font size in theme.json is medium,
so these blocks do not need the font size to be added in the pattern.

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/184

- Updating `hidden-search` pattern.
- Including the pattern in the different templates, as in Figma, they all use the same size.

**Screenshots**

https://github.com/user-attachments/assets/0e644a00-486c-444a-850e-9da236f1d748

**Testing Instructions**

1. Go to the site editor.
2. Look for the search results template.
3. Try every one of them, and confirm that the search section looks like in Figma.
4. Go to `yoursite.local/404` (replace `yoursite.local` with the site you're trying) and confirm that it also looks well on the 404 page.

### DIFF
--- a/patterns/hidden-404.php
+++ b/patterns/hidden-404.php
@@ -35,7 +35,7 @@
 				<!-- wp:paragraph -->
 				<p><?php echo esc_html_x( 'The page you are looking for doesn\'t exist, or it has been moved. Please try searching using the form below.', '404 error message', 'twentytwentyfive' ); ?></p>
 				<!-- /wp:paragraph -->
-				<!-- wp:search {"label":"<?php esc_html_e( 'Search', 'twentytwentyfive' ); ?>","showLabel":false,"placeholder":"<?php echo esc_attr_x( 'Type something...', 'input placeholder text', 'twentytwentyfive' ); ?>","buttonText":"<?php echo esc_attr_x( 'Search', 'button label', 'twentytwentyfive' ); ?>","style":{"border":{"radius":"999px"}}} /-->
+				<!-- wp:pattern {"slug":"twentytwentyfive/hidden-search"} /-->
 			</div>
 			<!-- /wp:group -->
 		</div>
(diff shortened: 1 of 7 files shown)


carolinan (changes requested):
Here is the difference between the pattern, and the search block:

https://github.com/user-attachments/assets/af8999e7-ed21-456e-941c-7f496872d3dd

In the pattern, the semi-transparent border is showing on the button,
I think that the border color should only be on the input. So it would need to be moved to the CSS in theme.json.

Since the default size for the block in theme.json is medium, the `fontSize` in the pattern markup should be removed.

juanfra:
Thanks, Carolina. Good catch. I've submitted the changes.

carolinan (approved):

PR #388, opened 2024-09-24
Add text styles

### LINKED ISSUE #305: Text styles

richtabor:
**Description**

An alternative to https://github.com/WordPress/twentytwentyfive/pull/330. 

Add text styles to alleviate the application of typographic values in patterns and templates. Closes https://github.com/WordPress/twentytwentyfive/issues/305. 

We'll need to follow-up, replacing `is-style-pill` style with `is-style-text-overline` throughout patterns, as well as using the other text styles in place of hard coded font size/line-height combinations throughout patterns. 

I changed "pill" to "overline" so that it's something more generic, making it more sustainable long-term. If a style variation wants to make them not pill shaped, it's no longer odd to choose the "Pill" style. 

**Screenshots**

CleanShot 2024-09-23 at 20 18 54

### DIFF
--- a/styles/blocks/01-title.json
+++ b/styles/blocks/01-title.json
@@ -5,7 +5,6 @@
 	"slug": "text-title",
 	"blockTypes": ["core/heading", "core/paragraph"],
 	"styles": {
-		"css": "display: inline-block;",
 		"typography": {
 			"fontSize": "var:preset|font-size|xx-large",
 			"lineHeight": "1.2"
--- a/styles/blocks/01-title.json
+++ b/styles/blocks/01-title.json
@@ -1,8 +1,8 @@
 {
 	"$schema": "https://schemas.wp.org/trunk/theme.json",
 	"version": 3,
-	"title": "Heading",
-	"slug": "text-heading",
+	"title": "Title",
+	"slug": "text-title",
 	"blockTypes": ["core/heading", "core/paragraph"],
 	"styles": {
 		"css": "display: inline-block;",
--- a/styles/blocks/02-heading.json
+++ b/styles/blocks/02-heading.json
@@ -1,8 +1,8 @@
 {
 	"$schema": "https://schemas.wp.org/trunk/theme.json",
 	"version": 3,
-	"title": "Title",
-	"slug": "text-title",
+	"title": "Heading",
+	"slug": "text-heading",
 	"blockTypes": ["core/heading", "core/paragraph"],
 	"styles": {
 		"typography": {
--- a/styles/blocks/02-subtitle.json
+++ b/styles/blocks/02-subtitle.json
@@ -1,8 +1,8 @@
 {
 	"$schema": "https://schemas.wp.org/trunk/theme.json",
 	"version": 3,
-	"title": "Heading",
-	"slug": "text-heading",
+	"title": "Subtitle",
+	"slug": "text-subtitle",
 	"blockTypes": ["core/heading", "core/paragraph"],
 	"styles": {
 		"typography": {
--- a/patterns/banner-about-book.php
+++ b/patterns/banner-about-book.php
@@ -27,8 +27,8 @@
 				<div class="wp-block-column" style="flex-basis:88%">
 					<!-- wp:group {"style":{"spacing":{"padding":{"bottom":"var:preset|spacing|50"}}},"layout":{"type":"constrained","contentSize":"500px","justifyContent":"left"}} -->
 					<div class="wp-block-group" style="padding-bottom:var(--wp--preset--spacing--50)">
-						<!-- wp:heading {"className":"is-style-text-heading"} -->
-						<h2 class="wp-block-heading is-style-text-heading">About the book</h2>
+						<!-- wp:heading {"className":"is-style-text-subtitle"} -->
+						<h2 class="wp-block-heading is-style-text-subtitle">About the book</h2>
 						<!-- /wp:heading -->
 
 						<!-- wp:paragraph {"fontSize":"medium"} -->
--- a/patterns/banner-about-book.php
+++ b/patterns/banner-about-book.php
@@ -27,8 +27,8 @@
 				<div class="wp-block-column" style="flex-basis:88%">
 					<!-- wp:group {"style":{"spacing":{"padding":{"bottom":"var:preset|spacing|50"}}},"layout":{"type":"constrained","contentSize":"500px","justifyContent":"left"}} -->
 					<div class="wp-block-group" style="padding-bottom:var(--wp--preset--spacing--50)">
-						<!-- wp:heading {"level":3,"className":"wp-block-heading","fontSize":"x-large"} -->
-						<h3 class="wp-block-heading has-x-large-font-size">About the book</h3>
+						<!-- wp:heading {"className":"is-style-text-heading"} -->
+						<h2 class="wp-block-heading is-style-text-heading">About the book</h2>
 						<!-- /wp:heading -->
 
 						<!-- wp:paragraph {"fontSize":"medium"} -->
--- a/patterns/banner-with-description-and-images-grid.php
+++ b/patterns/banner-with-description-and-images-grid.php
@@ -19,8 +19,8 @@
 		<div class="wp-block-group">
 			<!-- wp:group {"layout":{"type":"flex","orientation":"vertical","justifyContent":"left"}} -->
 			<div class="wp-block-group">
-				<!-- wp:heading {"className":"is-style-pill"} -->
-				<h2 class="wp-block-heading is-style-pill">About Us</h2>
+				<!-- wp:heading {"className":"is-style-text-subtitle"} -->
+				<h2 class="wp-block-heading is-style-text-subtitle">About Us</h2>
 				<!-- /wp:heading -->
 
 				<!-- wp:paragraph -->
--- a/patterns/banner-with-description-and-images-grid.php
+++ b/patterns/banner-with-description-and-images-grid.php
@@ -19,8 +19,8 @@
 		<div class="wp-block-group">
 			<!-- wp:group {"layout":{"type":"flex","orientation":"vertical","justifyContent":"left"}} -->
 			<div class="wp-block-group">
-				<!-- wp:heading {"className":"is-style-text-subtitle"} -->
-				<h2 class="wp-block-heading is-style-text-subtitle">About Us</h2>
+				<!-- wp:heading {"className":"is-style-text-overline"} -->
+				<h2 class="wp-block-heading is-style-text-overline">About Us</h2>
 				<!-- /wp:heading -->
 
 				<!-- wp:paragraph -->
--- a/patterns/contact-centered-social-link.php
+++ b/patterns/contact-centered-social-link.php
@@ -16,8 +16,9 @@
 <div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--80);padding-bottom:var(--wp--preset--spacing--80)">
 	<!-- wp:group {"align":"wide","layout":{"type":"flex","orientation":"vertical","justifyContent":"center"}} -->
 	<div class="wp-block-group alignwide">
-		<!-- wp:paragraph {"align":"center","style":{"typography":{"fontSize":"var(--wp--custom--font-size--huge)"}}} -->
-		<p class="has-text-align-center" style="font-size:var(--wp--custom--font-size--huge);"><?php echo wp_kses_post( _x( 'Got questions? <br><a href="#" rel="nofollow">Feel free to reach out.</a>', 'Heading of the Contact social link pattern', 'twentytwentyfive' ) ); ?></p>
+
+		<!-- wp:paragraph {"align":"center","className":"is-style-text-title"} -->
+		<p class="has-text-align-center is-style-text-title"><?php echo wp_kses_post( _x( 'Got questions? <br><a href="#" rel="nofollow">Feel free to reach out.</a>', 'Heading of the Contact social link pattern', 'twentytwentyfive' ) ); ?></p>
 		<!-- /wp:paragraph -->
 
 		<!-- wp:social-links {"iconColor":"contrast","className":"is-style-logos-only","layout":{"type":"flex","justifyContent":"center"}} -->
--- a/patterns/event-rsvp.php
+++ b/patterns/event-rsvp.php
@@ -24,8 +24,8 @@
 				<h1 class="wp-block-heading">Stories, historias, iсторії, iστορίες</h1>
 				<!-- /wp:heading -->
 
-				<!-- wp:paragraph {"fontSize":"x-large"} -->
-				<p class="has-x-large-font-size">Dec 12, 2024 — 10AM to 6PM</p>
+				<!-- wp:paragraph {"className":"is-style-text-title"} -->
+				<p class="is-style-text-title">Dec 12, 2024 — 10AM to 6PM</p>
 				<!-- /wp:paragraph -->
 
 				<!-- wp:spacer {"height":"0px","style":{"layout":{"selfStretch":"fixed","flexSize":"100px"}}} -->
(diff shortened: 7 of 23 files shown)


richtabor:
Here's an example using each of the text styles: 

CleanShot 2024-09-23 at 20 25 03

carolinan:
A  pull request and discussions about this already exist, and is already linked from the issue.
https://github.com/WordPress/twentytwentyfive/pull/330

richtabor:
My approach is slightly different :)

jasmussen:
Coming back to this after [the feedback I gave on the other PR](https://github.com/WordPress/twentytwentyfive/pull/330#issuecomment-2367782816), I have now a better understanding of what is desired here.

* We want at most 5 preset font sizes to maintain the ergonomic font size control.
* We ideally want some coherent and easily applicalble text-styles that can be applied across text blocks, that either leverage those same presets, or provide some additional nuance. 

One concern I shared also was the prominence of the style variation interface for applying these styles. It's huge. However I will say that the simpler Title / Heading / Subtitle labels help in this space a lot, both in communicating their purpose in the semantics of the document, but their utility as tools also.

Subtitle is the tricky one. I appreciate that it makes it more ergonomic to apply the pill-shape style, and to edit patterns that leverage this. It's also a pretty important element of the Twenty Twenty-Five design overall. At the same time, this is the only of the 4 variations that starts to fall apart once your paragraph becomes more than a single line. 

_Possibly_ this could be mitigated by changing the border radius to be a low number, such as 16px, then at least on very long paragraphs the curve wouldn't keep growing endlessly. The inline-block is also both fancy in terms of what it allows you to accomplish, yet it may be confusing when you're just formatting text. 

Would the patterns we need this for be possible without the inline-block part, and with a radius capped at 16px?

richtabor:
> Would the patterns we need this for be possible without the inline-block part, and with a radius capped at 16px?

I switched to inline-flex; seems to work better. And adapted the subtitle (pill) to better support multi-line. 

CleanShot 2024-09-24 at 10 09 01

jasmussen:
Okay, I'm happy to try this, especially given the looming beta deadline. Let's get this in and take stock.

jasmussen (approved):

juanfra (commented):
Thanks for working on this, Rich!

I think these slugs can be helpful to understand the nature of these styles, that's a big improvement. I'm not sold on "subtitle" though, as my expectation as a user would be different from what I'm getting (a bordered text style). But we can probably iterate on that.

What I wanted to ask, is if there's a reason behind choosing those sizes for the new text styles? For example, I see that "Heading" style is pretty much the same as "XL", and "Title" style is pretty much the same as "XXL".

"XL" and "XXL" are sizes that the user would be able to choose from the UI. And these new ones as well.

I think the problem that we're currently having with sizes and a11y (the one that was discussed [here](https://github.com/WordPress/twentytwentyfive/pull/286)) is that we don't have a way to select through UI sizes which are super close to H1 and H2. And I believe that if we make these styles look pretty close to those, we'll have that problem pretty much resolved.

We could set these new sizes to all patterns (H2 with "H1" sizes) and implementation will look almost the same as what we have in Figma. And "XL", "XXL" are going to still be there in case someone wants to use those sizes.

**Subtitle**

I see that the switch to `inline-flex` is causing some problems on some layouts.

https://github.com/user-attachments/assets/841e2f4f-50ad-487c-a6e4-5fd1f927e7c8

richtabor:
> What I wanted to ask, is if there's a reason behind choosing those sizes for the new text styles? For example, I see that "Heading" style is pretty much the same as "XL", and "Title" style is pretty much the same as "XXL".

I've applied the text style anywhere the XL and XXL values were used in combination with line height (1.2). 

> And "XL", "XXL" are going to still be there in case someone wants to use those sizes.

Exactly.

>  I'm not sold on "subtitle" though, as my expectation as a user would be different from what I'm getting (a bordered text style). But we can probably iterate on that.

Perhaps it comes down to nuance, but I can see subtitle as any sort of defined text to indicate an area, like the pill design we have, smaller text, a filled badge vibe, even all caps, etc. Picking a generic name allows users to define whatever "subtitle" is. It doesn't have to have a border etc.

styles/blocks/01-title.json
```diff
@@ -0,0 +1,13 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Title",
+	"slug": "text-title",
+	"blockTypes": ["core/heading", "core/paragraph"],
```
carolinan, line 6: I strongly disagree that a paragraph can be a "title". A paragraph should never look like a heading without being an actual heading, a heading needs to be semantic, it must use the correct HTML element.

carolinan: That is why I preferred the name agreed on in the other pull request, "Lead".

styles/blocks/02-heading.json
```diff
@@ -0,0 +1,13 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Heading",
+	"slug": "text-heading",
+	"blockTypes": ["core/heading", "core/paragraph"],
```
carolinan: Same here, absolutely no to a paragraph being styled as a heading.

carolinan: This would lead to users creating inaccessible content. If you have a paragraph, which is the default block, and there is a huge button at the top that says "heading" there is a big risk that users are going to believe that clicking on that button makes it a heading.

carolinan (changes requested):
The styles can not have names that indicate that paragraphs can in any way be used as headings and titles. The default theme encourages creating accessible content and sites.

Also, in some languages like Italian and Swedish, Heading and Title in the context of WordPress are translated into the same word.

carolinan:
If you are reading that I am bit frustrated it is because _we have actually already had this conversation_ in the TT5 group conversation on Slack in July.

richtabor:
I had asked for feedback on what to call that second style. Also, I removed the a11y label. It is not relevant.

richtabor:
I changed them to title, subtitle, and overline: 

CleanShot 2024-09-24 at 13 37 19

> Overline – A small, text style used above titles or sections to indicate a section of content.

styles/blocks/02-heading.json
richtabor: > Same here, absolutely no to a paragraph being styled as a heading.

I think you're missing the point. This is a text _style_. You can set text to whatever style, size, format you'd like—but the theme provides you defaults to pick from. Patterns are already doing this today, in this theme—they're just applied one off combinations of font size and line height. I am attempting to reduce the one-offs for a consistent systematized approach here.

richtabor: Also, I changed the name. I did ask for feedback on what to name this one as well.

styles/blocks/01-title.json
richtabor: It's a very common text style across many editing applications. I've never actually seen "Lead" used. We should not be recreating ideas and instead lead into the expectations users have, already set by many many other applications.

styles/blocks/02-subtitle.json
```diff
@@ -0,0 +1,13 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Subtitle",
+	"slug": "text-subtitle",
+	"blockTypes": ["core/heading", "core/paragraph"],
```
juanfra, line 6: We should probably also add `core/post-title` to this one and the "title", so we can use it in the context of the site editor. We could also consider the site title.

richtabor: Are you cool with us moving this in, then aligning on next steps to iterate on?

carolinan:
I still strongly disagree with these names, these proposals are dismissing the previous conversations and the work that has been made to avoid doing things like adding paragraphs **above** headings such as "overline".

styles/blocks/02-subtitle.json
juanfra: Yes, I believe that once this is merged, we could:

- Modify the `XXL` with the size of the H1.
- Modify the `XL` with the size of the H2.

The old XXL and XL will live on these new text styles. And then we can adjust [this one](https://github.com/WordPress/twentytwentyfive/pull/286) to use the modified XL and XL, instead of the custom variables.

styles/blocks/01-title.json
carolinan: What other applications do is not relevant **unless they produce accessible content**, and applying a text style to a paragraph to make it look like a heading means it will be perceived as a heading by sighted users, but not by search engines and screen reader users. Which is not the experience that users expects when using a default theme.

richtabor:
> I still strongly disagree with these names, these proposals are dismissing the previous conversations and the work that has been made to avoid doing things like adding paragraphs above headings such as "overline".

This does not change anything about the semantic structure of the page, but simply provides a systematic approach to typography.

styles/blocks/01-title.json
richtabor: I am not understanding the opposition to meet expectations. 

This is not about making headings designed like paragraphs, or paragraphs designed like headings. But rather to provide a set of typographic styles to be applied throughout, wherever necessary—whether that's with making a heading (any level) bigger, or paragraph smaller—it doesn't matter.

This is no different than making a paragraph have XL font size, and a line height of 1.2 (in trunk).

richtabor:
> I still strongly disagree with these names, these proposals are dismissing the previous conversations and the work that has been made to avoid doing things like adding paragraphs above headings such as "overline".

I'm concerned about the inconsistent use of heading levels within patterns. We should almost never (probably never) include H1 in most patterns. For example, why is this a `H1`?

We should assume H2 on almost all patterns, assuming levels from there (H3 when there are clearly other sub headings, etc) 

CleanShot 2024-09-24 at 15 22 44

styles/blocks/01-title.json
carolinan: I am not understanding the opposition to meet expectations of users that wants or has to create accessible content, and their visitors. 

The difference is that there is a big button at the top of the sidebar that encourages users to turn a paragraph into a heading, while not fulfilling that promise all the way. By only styling the text and not changing the HTML tag.
This is **encouraging** the wrong practice, compared to a text size option which does not encourage but also does not **prevent** wrong practices.

The words title and heading are translated into the same word, this already happens in the WordPress interface and editor today. Only using "title" as a style name and replacing "heading" resolves the duplication, but not the whole problem.

carolinan:
> This does not change anything about the semantic structure of the page, but simply provides a systematic approach to typography.

Exactly.
And having any element that is positioned like a heading and styled as a heading while not being a real heading is a problem.

carolinan:
I and others share the concern about the use of H1 in patterns. There are past discussions about this in separate issues, and ongoing work to review all patterns.

The only patterns that should have an H1 are:
 - Patterns intended to be full pages, with the page template that has no title.
- The first pattern on one of the landing pages.

Pages with these patterns must have one H1.

I have already brought up that in Twenty Twenty-Four, some designs like the RSVP was created as two patterns: One with the H1, meant to be a landing page, and one without an H1, meant to be placed inside content.
I asked wether this theme should do the same for some patterns, but there was no response and I intend to bring it up again as a separate issue for it to be considered, in case people just did not see the question.

jasmussen:
Noting there's strong disagreement, in catching up I would love to quickly try and zero in on what the main point of disagreement is, so we can reach a place where all are happy. From my vantage point, the main benefit of this particular approach is that the added type styles sound like _type styles_ rather than visual styles. I believe that's key for to maintain, if we are to move forward with these tools. If I'm right that the names are the main source of disagreement, let's brainstorm. Some ideas:

1. Title alternatives
* Headline
* Banner
* Display
* Primary
2. Subtitle
* Subheader
* Deck
* Subheading
* Secondary
* Lead
* Strapline
3. Overline
* Lead-in
* Tagline
* Preface
* Annotation

Any of those stand out?

styles/blocks/01-title.json
richtabor: > The difference is that there is a big button at the top of the sidebar that encourages users to turn a paragraph into a heading, while not fulfilling that promise all the way.

You missed my comments + the commits. I have changed it from "Heading". I had originally asked for feedback on what to name this, to not confuse it with headings. Did you see the commits?

richtabor:
I don't think the issue is naming conventions—but rather simply to provide mechanisms to apply text styles with. 

Today, in trunk, text styles are applied as combinations of typographic choices to an array of headings and paragraphs. This effort does not change that. Instead, it extracts those combinations of styles and systematizes the approach, with one source defining how each commonly used text style is styled.

This has nothing to do with accessibility semantics—you can use whatever heading levels are necessary for your content. It’s purely a design system approach to make styling content easier.

styles/blocks/01-title.json
carolinan: I did not miss the commit, and I wrote:

The words title and heading are translated into the same word, this already happens in the WordPress interface and editor today. Only using "title" as a style name and replacing "heading" resolves the duplication, but not the whole problem.

carolinan:
Subtitle is fine for me.
Title is fine if paragraph support is removed.
Overline is fine if paragraph support is removed.

carolinan:
I agree that pill is not a _perfect_ name, it was only initially used because it is the name of the existing shape for the social icons: a shape that users are familiar with in the editor interface, and because it is used for the matching style in Twenty Twenty-Four.

richtabor:
> Title is fine if paragraph support is removed.
> Overline is fine if paragraph support is removed.

That defeats the purpose of these text styles...

jasmussen:
Phew, this is a tricky one, and I'm wondering if there's a compromise we can find. 

* It is reasonable to suggest that type style names that might confuse you with the actual _semantic_ heading block, might cause people to use those variations instead of actually using the heading block.
* I see the sense in, if we're going to have 4 (including the default style) variations across all _text_ blocks, it makes sense to include them across all text blocks, including paragraph.
* There's a good argument for providing additional ergonomic shortcuts to providing typographic choices to lay out your content.

In that sense, and all other concerns aside, it does seem like there could be a way to satisfy all viewpoints by carefully finding the right names. If Overline suggest it can be placed before a _heading block_, and "Title" suggests "heading", then how would you all feel about the following, across _all_ text blocks?

* Default
* Display
* Subtitle
* Annotation

Here's how it would look:

image

Alternately, we could do "Epigraph". Here's how that would look:

image

Or a simpler word could be "Tagline":

image

There's a lot of nuance in this, but perhaps one of the above combinations can thread the needle?

carolinan:
Thank you! That summarizes it perfectly for me.
Of these suggested names I prefer the first.

I have never heard the word "Epigraph".
Tagline is too close to the site tagline block? If nobody else has that association, tagline is fine.

A small concern, not blocking, is that Display may be confused with using a so called "Display font"
But I am not able to come up with alternatives.

jasmussen:
A couple of quick additional ideas for an alternative to "Annotation": Note, Hook, Mark, Cue, Pointer. I'm not specifically suggesting these, just putting into place a couple of thoughts so we can hopefully balance things out.

carolinan:
During the Theme's team meeting on Slack, Justin Tadlock mentioned "lede" and "kicker" which were some of the ideas in the now closed PR.

jasmussen:
I considered those too, and can support especially "Lede" (which I think is also sometimes spelled "Lead" or "Leade"), though personally I find some of the other alternatives slightly clearer. But if Lede threads the needle here, I can go with it!

jasmussen:
I ended up trying #432 both to try the new phrases, and potentially to avoid the conflicts of this branch for the immediate future. Let me know if that works!

richtabor:
Closed in favor of https://github.com/WordPress/twentytwentyfive/pull/432, thanks all!

PR #273, opened 2024-09-11
Create CTA: grid with products and link pattern #68

graylaurenm:
**Description**

This pull request houses work on the pattern tracked in issue #68 

**Screenshots**

These are of the initial, incomplete version.
Screenshot 2024-09-11 at 12 32 25 PM
Screenshot 2024-09-11 at 12 32 45 PM
Screenshot 2024-09-11 at 12 32 53 PM

**Testing Instructions**

Insert this block into any page.

PR #390, opened 2024-09-24
 Create CTA: grid with products and link pattern, alt

jasmussen:
**Description**

Alternative to #273 and based on the work in https://github.com/WordPress/twentytwentyfive/pull/273#issuecomment-2370985814. Props @graylaurenm for the work.

**Screenshots**

370261170-0727539f-debb-47d5-915e-2fdff263dc5b


jasmussen:
Note that for some reason this pattern isn't showing up for me in the pattern inserter. Not sure what I'm doing wrong.

juanfra (commented):
Thanks for working on this, Joen!

I've tested this one and I see the pattern in the inserter. Although, the colored covers are looking smaller in that view and it doesn't represent the pattern perfectly. But that's likely unrelated to the PR. You can see what I referenced in the following screencast:

https://github.com/user-attachments/assets/235ab7ca-a08c-45c7-b5a5-359736bf442c

Regarding the pattern, everything looks good. The only thing I saw is that the grid items that only contain text are the only ones that are not looking squared on mobile. 

The other unrelated Gutenberg issue I found, is that squared images are not working well on Firefox.

jasmussen:
Thanks for pushing the change, and for reviewing. I wonder if we can't fix the issue with collapsing text by making those transparent cover blocks with 1:1 aspect ratio too. I have to go to a very late lunch now, in case you get to it, feel free to push such a change. Otherwise I can take a look later.

> The other unrelated Gutenberg issue I found, is that squared images are not working well on Firefox.

This is gutenberg core yes? Do you know if this is tracked already? Pretty sure that we're using web-standard `aspect-ratio: 1/1;` and `object-fit: cover;` to accomlish this, so hopefully it can be something to fix on the Firefox side, unless the block editor is doing something wrong.

juanfra:
> I wonder if we can't fix the issue with collapsing text by making those transparent cover blocks with 1:1 aspect ratio too

Yes, that's what I was thinking. I've just pushed changes to wrap those texts in a cover block (1:1) rather than stack. It's working well for me now.

> This is gutenberg core yes? Do you know if this is tracked already? Pretty sure that we're using web-standard aspect-ratio: 1/1; and object-fit: cover; to accomlish this, so hopefully it can be something to fix on the Firefox side, unless the block editor is doing something wrong.

Yes, it's Gutenberg core. I was checking this one out and it seems that Firefox needs to have the width and height on images for aspect-ratio/object-fit to work. I couldn't find any issue that's tracking this. `.wp-block-image img` already has a `max-width` of 100%; ~if we also add `width` to 100%, it should get resolved~ if we add a width of `fit-content` it should be resolved.


https://github.com/user-attachments/assets/f88ba7dc-acf7-4714-a735-76fd34fcf481

carolinan:
If you update the viewport width in the pattern file header to 1400 it will look like this in the inserter instead:
image

jasmussen:
Nice! Works well. Good to go? Thank you for the help, all.

juanfra:
That sounds good to me.

jasmussen:
🙏

PR #391, opened 2024-09-24
Deduplicate one image.

### LINKED ISSUE #314: Can the number of images be reduced?
Some of the images are included more than once but in different formats / sizes.
Would it be possible to use the aspect ratio or sizing in the patterns instead of using separate image files?

jasmussen:
**Description**

Fixes #314. Retires one image. Essentially deleting the small version of team-photo-3, and using the larger version in all patterns instead. Affected patterns, after this PR is applied:

Screenshot 2024-09-24 at 14 09 45

Screenshot 2024-09-24 at 14 10 11


carolinan (approved):
Thank you.

juanfra (approved):
Thanks, Joen! 🚀

PR #392, opened 2024-09-24
Update: Try fixing the text wrapping of the site title

### LINKED ISSUE #389: Footer Template: Site title wrapping
**Description**

The site title in the footer template may wrap unnaturally.

image

Change your site title. Results will vary depending on the length of the text, but the short site title will likely highlight this issue.

**Expected behavior**

There is enough space to the right of the site title that I would expect the text to not wrap:

image

**Environment info**

jasmussen:
**Description**

Fixes #389 (hopefully).

The site title and tagline uses a Columns block with two columns, the right-most has a spacer. Combined with "Stack on mobile", this causes that spacer block to provide a gap between title, tagline, and navigation items on mobile.

The two columns were 66% and 33% wide. So there wasn't a ton of room for the site title. This PR keeps the columns intact, but makes the first column 100% wide, with the second column not having a width definition, meaning it should expand, and notably keep its height, when stacking on mobile.

@juanfra let me know if this makes sense, is the right read of the intent, and if it helps fix the issue!

**Screenshots**

test

**Testing Instructions**

Test the default footer pattern responsively. The title should ideally not wrap.

### DIFF
--- a/patterns/footer.php
+++ b/patterns/footer.php
@@ -18,52 +18,53 @@
 	<div class="wp-block-group alignwide">
 		<!-- wp:site-logo /-->
 
-		<!-- wp:group {"align":"full","layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between","verticalAlignment":"top"}} -->
-		<div class="wp-block-group alignfull">
-			<!-- wp:columns -->
-			<div class="wp-block-columns">
-				<!-- wp:column {"width":"100%"} -->
-				<div class="wp-block-column" style="flex-basis:100%"><!-- wp:site-title {"level":2} /-->
+		<!-- wp:columns -->
+		<div class="wp-block-columns">
+			<!-- wp:column {"width":"40%"} -->
+			<div class="wp-block-column" style="flex-basis:40%">
+				<!-- wp:site-title {"level":2} /-->
 
 				<!-- wp:site-tagline /-->
-				</div>
-				<!-- /wp:column -->
-
-				<!-- wp:column {"width":""} -->
-				<div class="wp-block-column">
-					<!-- wp:spacer {"height":"var:preset|spacing|40","width":"0px"} -->
-					<div style="height:var(--wp--preset--spacing--40);width:0px" aria-hidden="true" class="wp-block-spacer"></div>
-					<!-- /wp:spacer -->
-				</div>
-				<!-- /wp:column -->
 			</div>
-			<!-- /wp:columns -->
+			<!-- /wp:column -->
 
-			<!-- wp:group {"style":{"spacing":{"blockGap":"var:preset|spacing|80"}},"layout":{"type":"flex","flexWrap":"wrap","verticalAlignment":"top","justifyContent":"space-between"}} -->
-			<div class="wp-block-group">
-				<!-- wp:navigation {"overlayMenu":"never","layout":{"type":"flex","orientation":"vertical"}} -->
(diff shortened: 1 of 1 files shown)


juanfra (commented):
Thanks for working on this Joen!

I've tested it and it works well on most scenarios. I've found that if the tagline is too long then the left column will take the full width of the footer and it'll look a bit broken.

I was taking a look at the footer structure and I see that the columns are wrapped in a row (which seems to be the culprit of the original issue) and I'm wondering if we really need it. If we get rid of that row we could possibly recreate the design and fix the wrapping issue.

https://github.com/user-attachments/assets/84fe4687-64b0-48a3-8080-99ae7425c7ad

jasmussen:
I think that's worth a shot. I may get time to work on this later today, but have to focus on some other things for the immediate time. If you have the code ready, feel free to push, otherwise I'll come back later and give it a shot.

jasmussen:
In a quick test of doing this, I couldn't replicate the design you ended up with. Notably we want the _navigation item_ to be fluid width on the right, but not stretched, same with the item on the left. Row plus space-between seems necessary to accomplish that, so not quite sure how you managed in your version. We want to avoid this, which I ended up with in retracing your steps:

Screenshot 2024-09-25 at 12 48 57

juanfra:
@jasmussen I've just pushed a quick test. I added a middle column with a spacer to make it more aligned with [the design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3056-190&t=9rSmqaKfi8N14IoO-1), including the limited space for the site tagline and the correct spacing for the navigation on the right. 

For the right part with the two navigations, I opted for a grid block, as it'll also be more in line with what we have in Figma. I also showed in the recording how it'd look if we kept a row there.


https://github.com/user-attachments/assets/ea14cc38-9b8b-4f96-ae53-3da6cab88bdf

Please let me know what you think, and feel free to revert if you don't believe it is a good direction.

jasmussen:
Hmm, I'm not too confident in the grid block, actually, I feel like there are still many rough edges to refine with that one. I also will admit that the not-fixed gap between the two columns of navigation items isn't great. I wonder if we should revert back to the initial setup, after all. Looking at your video here: https://github.com/WordPress/twentytwentyfive/pull/392#pullrequestreview-2325887804— the fact that when you add a full paragraph as a tagline (16s in the video) that actually works acceptably. Actually let me try a thing real quick.

jasmussen:
Hmm. The idea I wanted to try was to just justify-right the navigation items:

Screenshot 2024-09-25 at 13 34 02

But if we do that, they're also justified right on mobile, but we want this instead:

Screenshot 2024-09-25 at 13 35 02

So at the moment, unless we can set a max-width on the title+caption columns block, we should probably rewind.

jasmussen:
Okay, I rewinded to having the columns that causes the nav items to wrap if it's too long. I still think that's the best overall compromise, at least for now. There should still be time to fix issues like this after Friday, as they are bugs, but curious if you're okay with this for now @juanfra!

Note, I hope I got the rebase right. Works well in my testing.

juanfra (approved):
The rebase worked well for me. We're back at [here](https://github.com/WordPress/twentytwentyfive/pull/392#pullrequestreview-2325887804). Hopefully we can find a solution for that.

In the meantime, we can merge this to solve the site title wrapping issue.

PR #394, opened 2024-09-24
News blog templates: Add main element

carolinan:
**Description**

Adds the `<main>` HTML element to the news blog templates to improve the accessibility.

Partial for 
https://github.com/WordPress/twentytwentyfive/issues/278

**Testing Instructions**
Test all of the news blog templates:
- There should be no visual change on the front of the site
- All templates should have one `<main>` element
- All templates should have a "skip link":
 Place the curser in the address bar of the browser. Press the `Tab` key: move past the WordPress admin bar.
 Confirm that there is a skip link:
 
image


patterns/news-blog-single-post-with-sidebar-template.php
```diff
@@ -14,34 +14,39 @@
 ?>
 <!-- wp:template-part {"slug":"header","area":"header"} /-->
 
-<!-- wp:group {"layout":{"type":"constrained"}} -->
-<div class="wp-block-group">
-	<!-- wp:group {"align":"wide","layout":{"type":"default"}} -->
+<!-- wp:group {"tagName":"main","layout":{"type":"constrained"}} -->
+<main class="wp-block-group">
```
carolinan, line 18: I am not super happy about the amount of groups / divs in this template and it may be possible to simplify it without visual changes.

carolinan:
For the News blog with sidebar template, I would have liked to only use the `<main> ` around the posts, but because the sidebar is between the post sections, I decided not to.
I don't think that the second post section should be an _aside_ or _section_ outside the _main_ because it is an important part.

carolinan:
In "News blog with featured posts grid" I found that the newsletter signup pattern had been replaced with just the blocks, so I re-added the pattern.

juanfra (approved):
Thanks for working on this, Carolina! I've tested all and they work well, and with the respective "Skip to content" link.

carolinan:
And thank you for the review.

PR #396, opened 2024-09-24
Try: Use copyright in footer instead of `Twenty Twenty-Five` string.

### LINKED ISSUE #385: Discussion: Replace hardcoded `Twenty Twenty-Five` in footers.
**Description**

This is a follow-up of https://github.com/WordPress/twentytwentyfive/pull/372#issuecomment-2367817972.

We currently have three footers using `Twenty Twenty-Five` as a string, using the paragraph block. We're not using the site title or anything dynamic. As a result, any user who won't like having the theme name in the footer, will have to manually and purposefully edit the footer to change it.

- [patterns/footer.php](https://github.com/WordPress/twentytwentyfive/blob/trunk/patterns/footer.php#L75)
- [patterns/footer-columns.php](https://github.com/WordPress/twentytwentyfive/blob/trunk/patterns/footer.php#L75)
- [patterns/footer-newsletter.php](https://github.com/WordPress/twentytwentyfive/blob/trunk/patterns/footer-newsletter.php#L45)

Having the site title where we now have `Twenty Twenty-Five` can be redundant in some of these footers (default and columns), as it's already being used. 

A good candidate to fill this space could be the [copyright pattern](https://github.com/WordPress/twentytwentyfive/blob/trunk/patterns/copyright.php) we're introducing with the block binding that will print `© YEAR`.

Screenshot 2024-09-23 at 15 32 55

In the screenshot, `Stories®` is the site title, and `The new WordPress default theme pays homage to history and ancestry.` the tagline.

### LINKED ISSUE #399: Copyright "pattern" should not be a pattern
Let's be a bit more opinionated on what constitutes a pattern. A copyright text is not an ideal component for building out pages with—we should remove it as a standalone pattern and instead use it within patterns/templates that make sense. 

CleanShot 2024-09-24 at 16 39 12
CleanShot 2024-09-24 at 16 39 15

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/385

- Use the copyright pattern instead of hardcoding `Twenty Twenty-Five`
- Remove the text color from the pattern.
- Wrap the pattern in a group with text color for the "Portfolio Homepage" to match the designs.

**Screenshots**

https://github.com/user-attachments/assets/d2d832f4-29c5-43a6-8b62-4d1e85ca5bff

**Testing Instructions**

Copyright in footer

1. On the site editor, edit the footer.
2. Confirm that the three footers described in the issue are using the copyright pattern instead of the `Twenty Twenty-Five` hardcoded string.

Copyright in "Portfolio Homepage"

1. Create a page.
2. Use the "Portfolio Homepage"
3. Confirm the copyright pattern is still using the same text color.


jasmussen:
I could go with this. @richtabor, any thoughts?

richtabor:
I think it's better to **not** have a copyright "pattern", but just put the block in the actual footers. It's unnecessary complexity, and kinda breaks the notion of using patterns to build out a page. Related: https://github.com/WordPress/twentytwentyfive/issues/399

carolinan:
I agree that the copyright pattern could be hidden from the inserter -unless there is a way to only make it available when editing footers?

But I don't think it creates complexity. I think it reduces complexity, I see no benefit in repeating the code in multiple places
when it can be managed from one.
The code block for the block binding metadata in the copyright pattern is relatively large, but I would _prefer_ to keep it readable, since it is a new feature.

richtabor:
> But I don't think it creates complexity. I think it reduces complexity, I see no benefit in repeating the code in multiple places
when it can be managed from one.

It does though, you'll have to find the pattern in the code—tracing it down. It's not more complicated than a paragraph block with a few attributes. This makes it seem like you need to orient it like the theme does (a standalone pattern), which is more complicated than just putting it wherever necessary. 

CleanShot 2024-09-25 at 10 25 47

 It can also be reduced to not have color, or the 'copyright' class on it (is that used for anything?): 

```
<!-- wp:paragraph { "metadata":{ "bindings":{ "content":{ "source":"twentytwentyfive/copyright" } } }, "fontSize":"small" } -->
<p class="has-small-font-size"></p>
<!-- /wp:paragraph -->
```

carolinan:
The copyright class is only a utility similar to how many block themes include the classes for sidebar, site-footer and site-header.
The color has already been removed I believe.
If the class name really brings no value to users, yes, let's remove the pattern :)

juanfra:
I've just pushed changes to update the code and remove the pattern. Once merged, this one should fix https://github.com/WordPress/twentytwentyfive/issues/399 and https://github.com/WordPress/twentytwentyfive/issues/385

PR #400, opened 2024-09-24
Replace fixed elements with a placeholder

eliorivero:
**Description**

This solves the issue where the **Portfolio Homepage** template was showing multiple "No posts were found." messages. By using the placeholder we can show this message only in the editor and avoid displaying it on the front end.

**Screenshots**

Used two languages, Spanish and German, to better highlight what's text produced by WordPress (the localized text) and which one was the fixed text from the pattern (text in English).

### Previously, the editor was showing multiple "No posts were found."
Screenshot 2024-09-24 at 23 35 04

### Which were also visible on the front end
Screenshot 2024-09-24 at 23 36 18

### With this approach, the messages are shown as placeholders in the editor
Screenshot 2024-09-25 at 00 42 36

### and the messages aren't visible on the front end
Screenshot 2024-09-25 at 00 30 04

**Testing Instructions**

1. Activate the theme.
2. Add a page.
3. Select the Portfolio Homepage pattern.
    Screenshot 2024-09-25 at 00 46 45
4. Verify the placeholders are there.
5. Publish the page and check on the front end that the placeholders aren't visible.

### DIFF
--- a/patterns/page-portfolio-home.php
+++ b/patterns/page-portfolio-home.php
@@ -58,8 +58,7 @@
 					<!-- /wp:post-template -->
 
 					<!-- wp:query-no-results -->
-					<!-- wp:paragraph -->
-					<p>No posts were found.</p>
+					<!-- wp:paragraph {"placeholder":"Add text or blocks that will display when a query returns no results.","fontSize":"medium"} -->
 					<!-- /wp:paragraph -->
 					<!-- /wp:query-no-results -->
 				</div>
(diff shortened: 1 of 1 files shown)


carolinan:
Question: what is the benefit of keeping the blocks with the placeholders in the editor?

I think it is best to keep the message on the first query visible on the front. 🤔 Otherwise the page will say "...these are some..." and then there is nothing, without an explanation.

eliorivero:
IMHO the benefit is that the users are made aware that there's a position there to insert content.

carolinan:
@eliorivero Thank you for working on this. Since there is no consensus on if these should be changed or not, and because the time until the release is short, I am going to close this PR.

PR #414, opened 2024-09-25
Simplify button and link styles

### LINKED ISSUE #407: Default variation: Buttons have the wrong styles
**Description**

In the default style variation, the filled buttons have the wrong hover style both outside a section and when placed in a styled section. They have the wrong border color.

The outline variation has the wrong border color in the default state.

**Expected behavior**

For the design to match Figma: 
https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=77-2003&t=FdIAd6LHSrvjzxYG-4

https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=565-10053&t=4uwE2wefr81mwKEU-4

**Screenshots**
In this clip I am hovering over the buttons in section style 1 and 2.

https://github.com/user-attachments/assets/069f037f-c684-48bd-aa53-3ddf705df4be

richtabor:
**Description**

Simplifies button and link styles, reducing conflicts between block style variations (pattern/section styles). 

- Reduce custom code in button-outline.css (we should consider moving this to theme.json)
- Reduce dependencies on setting link color in section styles, as parent theme.json uses currentColor. This makes it easier to change colors of a section without having to also change its link colors. 
- Remove the hover effect applying accent-1, instead using a color-mix of the button's background color. This reduces the mini conflicts with style variations.
- Make button styles for the outline variation match visually to the default.  
 
**Screenshots**

CleanShot 2024-09-25 at 15 37 57

Closes https://github.com/WordPress/twentytwentyfive/issues/407.

### DIFF
--- a/assets/css/button-outline.css
+++ b/assets/css/button-outline.css
@@ -1,6 +1,3 @@
-.wp-block-button.is-style-outline
-	> .wp-block-button__link:not(.has-text-color, .has-background):hover {
-	background-color: var(--wp--preset--color--contrast, transparent);
-	color: var(--wp--preset--color--base);
-	border-color: var(--wp--preset--color--contrast, currentColor);
+.wp-block-button.is-style-outline > .wp-block-button__link:not(.has-background):hover {
+	background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 5%, transparent);
 }
--- a/styles/blocks/section-1.json
+++ b/styles/blocks/section-1.json
@@ -19,19 +19,6 @@
 					"text": "color-mix(in srgb, currentColor 25%, transparent)"
 				}
 			}
-		},
-		"elements": {
-			"button": {
-				"color": {
-					"background": "var:preset|color|contrast",
-					"text": "var:preset|color|base"
-				}
-			},
-			"link": {
-				"color": {
-					"text": "var:preset|color|contrast"
-				}
-			}
 		}
 	}
 }
--- a/styles/blocks/section-2.json
+++ b/styles/blocks/section-2.json
@@ -23,13 +23,14 @@
 		"elements": {
 			"button": {
 				"color": {
-					"background": "var:preset|color|contrast",
-					"text": "var:preset|color|base"
-				}
-			},
-			"link": {
-				"color": {
-					"text": "var:preset|color|accent-2"
+					"background": "var:preset|color|accent-2",
+					"text": "var:preset|color|accent-3"
+				},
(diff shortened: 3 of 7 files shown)


carolinan:
I had been looking forward to exploring and cleaning this up which is why I had assigned it to myself and started working on it.

jasmussen (approved):
This is working well for me. Things look good and work well in the editor:

editor

The hover styles are nice:

after, frontend

Colors also work:

colors

I'd love a quick look from @carolinan to ensure details from #407 are captured.

juanfra (approved):
Thanks, Rich! Looking good and working well for me.

https://github.com/user-attachments/assets/d4b0a56e-785c-4215-9d42-6136c7414022

I'll leave a ping for @beafialho for her to review once she's back, as the hover styles differ from what she originally posted [here](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=2431-1113&t=DWoVwZDlj9tFQiL2-1).

carolinan:
Tools like the Firefox built-in contrast checker and browser extensions are not able to determine the color contrast automatically because of the color-mix. With some tools you may see a result, but this result will not be accurate.

As long as there is an understanding that the colors **must be accessible**, and that this may require changes, it can be updated in a separate pull request, But this must be a team effort. Checking this will require manual testing which I can not do on my own because of the amount.
Both the button elements and the button blocks must be tested.

Are you able to each pick a style section and test the button states across the different theme style variations?
Or maybe you would prefer to take a few theme style variation each and test all section styles.

I am going to create a separate issue for the `:focus` style, because while this PR is a regression; it was not accessible before this PR, either.

carolinan:
For the :hover state, that the cursor changes can be considered enough to indicate that an item is being hovered.
It would be the absolute minimum and the theme aims to do better.
What to test: 
On the button block, check for a high enough color contrast ratio between the text color and background color when the item is hovered.

### Example
Default theme style variation
- [x] All passes 🎉 In section 2, the filled button has a text to background contrast ratio of 4.84:1 where 4.5:1 is the minimum.

Evening
The hover on the filled button is broken _or not implemented?_

Noon
- [x] Button color passes 🎉 

In two style sections, both headings and links are invisible.

Dusk
This was the first variation added, so it needs a clean up of both button and link colors.

jasmussen:
Thanks, all!

PR #420, opened 2024-09-25
Rename primary/secondary/opacity colors to accents

### LINKED ISSUE #415: Change primary and secondary color slugs to "accent" colors
Do "primary" and "secondary" have any significance? It seems we could just make them more accent colors, instead of insinuating importance for these. As-is, it's not clear what's different between these and "accent" colors.  

The opacity color should really be an accent color as well. 

CleanShot 2024-09-25 at 16 07 42

EvanHerman:
Resolves #415

**Description**

As mentioned in the original issue, it is not clear what significance the `Primary`, `Secondary` and `Opacity` colors have and should actually be accent colors. This PR changes the `Primary` color to `Accent 4`, `Secondary` to `Accent 5` and `Opacity` to `Accent 6`.

**Screenshots**

image
image
image

**Testing Instructions**

- Edit the global styles.
- Click on 'Palette'.
- Hover over each color to see their name appear in a tooltip.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -35,18 +35,18 @@
 				},
 				{
 					"color": "#686868",
-					"name": "Primary",
-					"slug": "primary"
+					"name": "Accent 4",
+					"slug": "accent-4"
 				},
 				{
 					"color": "#FBFAF3",
-					"name": "Secondary",
-					"slug": "secondary"
+					"name": "Accent 5",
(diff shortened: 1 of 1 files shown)


richtabor:
@EvanHerman we'll need to do the same in the style variations within /styles/*

PR #421, opened 2024-09-25
Add better core/columns spacing blockGap default

richtabor:
Instead of having to define them within every pattern, just define the value in theme.json instead. 

### Before

CleanShot 2024-09-25 at 17 10 24

### After 

CleanShot 2024-09-25 at 17 10 10

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -995,6 +995,14 @@
 					}
 				}
 			},
+			"core/columns": {
+				"spacing": {
+					"blockGap": {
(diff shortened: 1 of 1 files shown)


jasmussen (approved):

carolinan (approved):
This works great, thank you.

carolinan:
Are we sure this is valid? Because now I am seeing: Incorrect type, expected string.

image

carolinan:
Is it a schema version issue?

PR #397, opened 2024-09-24
Removing extraneous padding left and right from patterns

richtabor:
Improvements centered on consistency across patterns. 
- Removes unnecessary left/right padding values. 
- Improves mobile view of many patterns. 
- Reduces complexity of many patterns by flattening unnecessary groups etc. 
- Fix the big "Stories" text to not grow past the viewport on big screens. 
- Start organizing patterns into proper categories. 

CleanShot 2024-09-24 at 16 51 39
CleanShot 2024-09-24 at 17 04 50

CleanShot 2024-09-24 at 16 51 18
CleanShot 2024-09-24 at 16 51 10
CleanShot 2024-09-24 at 16 51 04
CleanShot 2024-09-24 at 16 50 55
CleanShot 2024-09-24 at 17 04 46

PR #424, opened 2024-09-26
Vertical header: Fix block validation error

carolinan:
**Description**
When the padding setting was removed from the pattern in https://github.com/WordPress/twentytwentyfive/pull/397,
the CSS was left inside the style attribute, which caused a block validation error.

This PR removes the CSS.

**Testing Instructions**
View the vertical header in the editor and confirm there is no block validation error.


juanfra (approved):
Thanks Carolina! I don't see any block validation error, it works well.

https://github.com/user-attachments/assets/3873f6b8-a625-442c-b9dd-010d763bea9d

Issue #395, opened 2024-09-24
Reduce the number of "No posts were found" messages


carolinan:
**Is your feature request related to a problem? Please describe.**

On my test installation I have 3 posts, which means that on the News Blog home template there are  four "No posts were found" messages showing:
![image](https://github.com/user-attachments/assets/ef8be6f4-2565-42e9-8e39-9a6beae70401)

In my _opinion_ this feels too repetitive.

**Describe the solution you'd like**

Only show one "No Posts were found" message in this template.

carolinan:
Related: https://github.com/WordPress/twentytwentyfive/pull/400

jasmussen:
Is this addressed as of #427?

carolinan:
In the templates that have a large number of queries, the number of visible messages are still the same yes?
In https://github.com/WordPress/twentytwentyfive/pull/400 @eliorivero had the idea to keep the blocks, but most of them would be placeholders.

My idea was to only show it on the front for the first query. If there are not enough posts in the first query, there are no posts to show in the following, either.

jasmussen:
If we get no other improvements in, hopefully at least it will be an edgecase or encouragement to add some posts in those spaces.

carolinan:
I don't feel that there is consensus on this, so let's not make any additional changes.
The user can delete the blocks too, if they need.

PR #427, opened 2024-09-26
Update: Improve "No results" order, style, and phrasing.

jasmussen:
**Description**

Partially related to #395, 

When opening the default blog home template in the site editor, the first thing you see is "Blog", followed by "No posts found."

before

This can easily be misunderstood as an error, especially since posts on your site are right there, showing up below.

What's going on behind the scenes—and this is worth improving in the block editor at the soure itself—is that the "No results" block has to be included in the query loop so that you can style it. but on the frontend, it will only actually show up if you really do have no results to show.

This PR does a few small things:

* It moves the "No results" section _below_ the posts. This mitigates the feeling of this being an error, and will work the same since either section will be hidden depending on the query.
* It rephrases the text to be slightly friendlier.
* It adds the same top/bottom padding to its Group container, as the post itself has, so that there's a consistent gap between title, post|no results, and pagination.

Result:

after

**Testing Instructions**

Test Blog home (and other archive templates) in the site editor, observe the new message showing up below the list of posts.

### DIFF
--- a/patterns/posts-personal-blog.php
+++ b/patterns/posts-personal-blog.php
@@ -28,7 +28,7 @@
 	<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)">
 		<!-- wp:query-no-results -->
 		<!-- wp:paragraph -->
-		<p><?php echo esc_html_x( 'Sorry, but nothing matched your search terms. Please try again with some different keywords.', 'Message explaining that there are no results returned from a search', 'twentytwentyfive' ); ?></p>
+		<p><?php echo esc_html_x( 'Sorry, but nothing was found. Please try a search with different keywords.', 'Message explaining that there are no results returned from a search', 'twentytwentyfive' ); ?></p>
 		<!-- /wp:paragraph -->
 		<!-- /wp:query-no-results -->
 	</div>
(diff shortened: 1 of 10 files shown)


patterns/posts-personal-blog.php
```diff
@@ -33,6 +24,15 @@
 		</div>
 		<!-- /wp:group -->
 	<!-- /wp:post-template -->
+	<!-- wp:group {"style":{"spacing":{"padding":{"top":"var:preset|spacing|60","bottom":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group" style="padding-top:var(--wp--preset--spacing--60);padding-bottom:var(--wp--preset--spacing--60)">
+		<!-- wp:query-no-results -->
+		<!-- wp:paragraph -->
+		<p><?php echo esc_html_x( 'Sorry, but nothing matched your search terms. Please try again with some different keywords.', 'Message explaining that there are no results returned from a search', 'twentytwentyfive' ); ?></p>
```
carolinan: I am not _searching for a term_ though 🤔 I am viewing a blog.

jasmussen: Good point. I mostly went this route because the same results pattern is used on the search results page. That is _presumably_ also where you'll see this message, in practice, most of the time.

How about:

"Sorry, but nothing was found. Please try searching again."

jasmussen: Or even, 

"Sorry, but nothing was found. Please try a search with different keywords."

carolinan: Yes that would work a bit better on all scenarios.

carolinan: This is just one template though. Some of the templates have a large amount of no results blocks:
https://github.com/WordPress/twentytwentyfive/pull/400

jasmussen: I'll update the phrase and see if I can update the remaining templates after lunch.

jasmussen:
Okay I took a stab. I've intentionally split the changes into two commits, because the latter was a big search and replace. Let me know if you think I got that right, and if not, I can revert that commit.

jasmussen:
Button is green. Is it okay to merge this one? Thank you for reviews!

PR #432, opened 2024-09-26
Alt: Just the new text-styles

jasmussen:
**Description**

Alternative to #330 and #388. 

Adds two new text styles, Display and Subtitle, and renames "Pill" to "Annotation", as discussed [here](https://github.com/WordPress/twentytwentyfive/pull/388#issuecomment-2376311686). Happy to update "Annotation" further, depending on your feedback.

The reason this is a separate PR is that this includes _only_ the .json files, and no pattern changes. We can explore those changes separately depending on first landing this. 

Penny for your thoughts! 🙏

### DIFF
--- a/styles/blocks/01-display.json
+++ b/styles/blocks/01-display.json
@@ -1,0 +1,13 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Display",
+	"slug": "text-display",
+	"blockTypes": ["core/heading", "core/paragraph"],
+	"styles": {
+		"typography": {
+			"fontSize": "var:preset|font-size|xx-large",
+			"lineHeight": "1.2"
+		}
+	}
+}
--- a/styles/blocks/02-subtitle.json
+++ b/styles/blocks/02-subtitle.json
@@ -1,0 +1,13 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Subtitle",
+	"slug": "text-subtitle",
+	"blockTypes": ["core/heading", "core/paragraph"],
+	"styles": {
+		"typography": {
+			"fontSize": "var:preset|font-size|x-large",
+			"lineHeight": "1.2"
+		}
+	}
+}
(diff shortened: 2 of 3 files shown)


juanfra:
Thanks, Joen! 

I like the names, thank you for coming up with [the alternatives](https://github.com/WordPress/twentytwentyfive/pull/388#issuecomment-2373437931). The only one that, I personally think could be different is "Display". It's subjective, but I'm biased because that wouldn't translate well or have a clear representation in Spanish.

From the alternatives you shared, I prefer "Primary" and "Headline" instead of "Display". The only caveat for Primary is that we already use it for a color in the color palette, and maybe the expectation from some users is that the text style has indeed the primary color.

Annotation and overline work well, in my opinion. I don't have a particular preference for that one; both work well.

carolinan:
To me, headline has an even stronger association to being the most important heading on a page than both heading and title, and I don't think it should be used to style a paragraph.

richtabor:
> Adds two new text styles, Display and Subtitle, and renames "Pill" to "Annotation", as discussed https://github.com/WordPress/twentytwentyfive/pull/388#issuecomment-2376311686. Happy to update "Annotation" further, depending on your feedback.

I'm cool with what @joen has proposed here. Let's move this forward, then perhaps circle back to adjust. It's critical to get type into a system that works, then we can rename to whatever we feel is best afterwards.

styles/blocks/01-display.json
```diff
@@ -0,0 +1,13 @@
+{
+	"$schema": "https://schemas.wp.org/trunk/theme.json",
+	"version": 3,
+	"title": "Display",
+	"slug": "text-display",
+	"blockTypes": ["core/heading", "core/paragraph"],
+	"styles": {
+		"typography": {
+			"fontSize": "var:preset|font-size|xx-large",
```
richtabor, line 9: noting these sizes may have to be one-off sizes once we pull this into https://github.com/WordPress/twentytwentyfive/pull/422 and evaluate properly.

richtabor: for both display and subtitle

richtabor (approved):

jasmussen:
Okay, I checked with Juanfra. Let's merge this and then revisit if it becomes necessary.

juanfra (approved):

PR #434, opened 2024-09-26
Add: Business & shop homepages

juanfra:
**Description**

Adding

- [Business home](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=1-3406&t=DWoVwZDlj9tFQiL2-1)
- [Shop home](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=1-3400&t=DWoVwZDlj9tFQiL2-1)

**Screenshots**

https://github.com/user-attachments/assets/093a1e84-4843-423d-9e4d-71f003faf3a3

**Testing Instructions**

1. Create a page.
2. See the new pages are in the modal.
3. Insert both and confirm they use the design's same pattern compilation.

### DIFF
--- a/patterns/page-business-home.php
+++ b/patterns/page-business-home.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * Title: Business Homepage
+ * Slug: twentytwentyfive/page-business-home
+ * Categories: twentytwentyfive_page, posts, featured
+ * Keywords: starter
+ * Block Types: core/post-content
+ * Post Types: page, wp_template
+ * Viewport width: 1400
+ * Description: A business homepage pattern.
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
+?>
+<!-- wp:group {"align":"full","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull">
+	<!-- wp:pattern {"slug":"twentytwentyfive/cta-centered-heading"} /-->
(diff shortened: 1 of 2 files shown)


jasmussen (approved):
Looks good. I'll defer to Carolina if she has feedback otherwise.

carolinan:
I did not realize that there were combined pages left to add.

Shop - The first heading is the "New arrivals", but it does not use an H1, so the landing page is missing the H1.
The problem is that this pattern can be inserted anywhere, where the H1 is not suitable, and we are again faced with the question of either
- hiding the pattern from the inserter,
- using two patterns, one for the landing page and one for the content.
- add an existing pattern that uses an H1, above this.


Business -I really like this combination.
The "our service" pattern is skipping a heading level. There is already a PR's queued to update this, but it is blocked by the font size discussion. If you are going to make any further changes to this PR, it can be updated here, or you can skip it.

juanfra:
Thanks for the input, Carolina.

> hiding the pattern from the being inserted,

It could be, but then we'll have one pattern less in the bundle. I believe it'd be valuable if this one was available for users.

> using two patterns, one for the landing page and one for the content.

I would prefer if we don't do this. It could be difficult to maintain and I don't think it's a viable alternative.

> add an existing pattern that uses an H1, above this.

That could be a good alternative. Although I believe Bea or Joen should be defining which one.

> The "our service" pattern is skipping a heading level. There is already a PR's queued to update this, but it is blocked by the font size discussion. 

👍

Given the timeline and the beta, I'd prefer if we add this before Friday's EOD and then iterate via follow-up issue on the a11y concerns.

PR #440, opened 2024-09-27
Try: Update the order of the section styles

### LINKED ISSUE #142: Add section styles

carolinan:
**Description**
I updated the order of the section styles from light to dark.
I placed the files in a new folder called sections.

I did not change the names, they are still called Style 1-5.

**Screenshots**
image

Before:
image

**Testing Instructions**

Add a group or columns block and confirm that the style sections have the correct color, but a different name and order than before the PR.
Add a few patterns that have a style section, and confirm that the color is the same as before the PR.

Contact location and link -yellow
Link in bio heading, paragraph, links and full-height image -purple
Hero podcast -pink
Logos -off white/ beige


styles/01-evening.json
```diff
@@ -119,7 +119,7 @@
 			}
 		},
 		"variations": {
-			"section-1": {
```
juanfra, line 122: Do we want to order the sections by number in the different style variations?

juanfra (commented):
Looking good and working well to me. I left a small comment (nit) about ordering the sections by number in the different style variations, like they were before. 

https://github.com/user-attachments/assets/b72cfc28-9f1a-42af-a14c-04ef76f03281

From my perspective, the order matches what @richtabor mentioned [here](https://github.com/WordPress/twentytwentyfive/issues/142#issuecomment-2378164800) (lightest to darkest), so that's good.

styles/01-evening.json
carolinan: Oh I did not think of that. I will make an update in the afternoon.

carolinan: I did not find the time yesterday, trying it now.

carolinan:
Evening:
The colors are still not sorted from light to dark but dark to light:
image

Morning: is also kind of OK
image

It gets worse from there:

Noon:
image

Dusk:
image

Afternoon:
image

Twilight:
image

Sunrise:
image

Midnight:
image

PR #441, opened 2024-09-27
Add patterns for post formats Link and Audio

jasmussen:
**Description**

Along with a #295, maybe this one helps address #17.

This PR adds two new patterns, Link format and Quote format. The motivation is this:

* Post format archives already exist, i.e. if you go to /type/quote you will see quote-based post formats.
* In a large amount of the post formats registered in #295 (standard, quote, video, gallery), the blocks with their default styles represent a post format styling, both on individual posts and in archive pages.
* The underlying templates to present post formats with a unique appearance is still lacking in places, so perhaps it's not worth having dedicated archive templates that change the apperance from what the post format will look like on its permalink. 

Specifically the Link and Audio formats, though, are harder to get visually right. This PR creates a shortcut for you, a few patterns. It's not clear whether this is a good idea or not, but if it's an idea to try, the PR would need to land soon.

Figma design for Link format:

figma-link

Corresponding pattern design:

link format

Figma design for audio:

figma-audio

Corresponding pattern design:

audio format

Noting the discrepancies above, we cannot style the audio player.

Separately, I tried categorizing these as post formats, and featured, though something here is missing, and I'm not actually seeing these show up. Let me know my mistake here!


carolinan:
> Post format archives already exist, i.e. if you go to /type/quote you will see quote-based post formats.

Yes it uses archive.html

carolinan:
Custom categories need to be registered: https://github.com/WordPress/twentytwentyfive/blob/trunk/functions.php#L113

carolinan:
For the audio, is it not possible to make the image the same height as the other content?

jasmussen:
I'll take a look at registering that cateogory. We can always remove it if we think it's a bad idea.

For the audio, I can probably do something with a Stack block, but in this case I wanted to try and keep the perfect _square_ shape of the image, which feels standard for podcasts. What do you think?

jasmussen:
Something's not working with my code, would appreciate a gut-check. Feel free to push if I'm missing something obvious. For now I moved these to the "Pages" category.

carolinan:
I see that the category is fixed.

carolinan:
Thank you :)

carolinan:
Oh the Stories link pattern has the registered trademark symbol, it needs to be removed.

jasmussen:
> Oh the Stories link pattern has the registered trademark symbol, it needs to be removed.

On it.

PR #442, opened 2024-09-27
Update pattern filenames, and titles

### LINKED ISSUE #244: Audit patterns filenames
**Description**

Our goal is to make the naming as intuitive as possible. To achieve this, we need to audit the patterns' names and define proper naming for the different templates in the `patterns` folder. We can check how it was done for the past default themes with patterns so that users are familiar with how we name things out of the box.

I believe we should consider the convention that was historically established in WordPress themes, but as prefixes: `single-*`, `page-*`, `archive-*`, `home-*`, `search-*`. Thinking also of `template-*` as an option.

We also need to find some convention for the other isolated patterns the theme is packed with, an idea could be thinking of the main category as the prefix.

jasmussen:
**Description**

Starts work at, maybe fixes #244. This PR does three things:

1. It renames pattern filenames according to [the proposed format here](https://github.com/WordPress/twentytwentyfive/issues/244#issuecomment-2377605175).
2. It updates pattern _slugs_ to reference match the filenames, and updates any includes to point to the new slugs.
3. It updates titles to use numbers instead of text (3 columns instead of three columns), sentence case instead of titlecase, and comma instead of a colon for prefixes.

This needs a good bit of testing to see that patterns work and notably item 2, to make sure I've captured every template that includes other patterns with the correct slug.

Feel free to push changes to this PR.


patterns/page-business-home.php
```diff
@@ -18,9 +18,9 @@
...
 <div class="wp-block-group alignfull">
 	<!-- wp:pattern {"slug":"twentytwentyfive/cta-centered-heading"} /-->
-	<!-- wp:pattern {"slug":"twentytwentyfive/overlap-images-and-paragraph"} /-->
-	<!-- wp:pattern {"slug":"twentytwentyfive/services-three-columns"} /-->
-	<!-- wp:pattern {"slug":"twentytwentyfive/testimonials-review-large-image-right"} /-->
+	<!-- wp:pattern {"slug":"twentytwentyfive/overlapped-images"} /-->
+	<!-- wp:pattern {"slug":"twentytwentyfive/services-3-col"} /-->
+	<!-- wp:pattern {"slug":"twentytwentyfive/testimonials-large"} /-->
 	<!-- wp:pattern {"slug":"twentytwentyfive/pricing-2-col"} /-->
 	<!-- wp:pattern {"slug":"twentytwentyfive/newsletter-sign-up"} /-->
```
juanfra:
```suggestion
	<!-- wp:pattern {"slug":"twentytwentyfive/cta-newsletter"} /-->
```

patterns/page-landing-book.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Title: Landing page for Book
+ * Title: Landing page for book
  * Slug: twentytwentyfive/page-landing-book
  * Categories: twentytwentyfive_page, featured
  * Keywords: starter
```
juanfra, line 6: In this one we also need to update the `newsletter-signup` to `cta-newsletter`

patterns/template-search-news-blog.php
```diff
@@ -30,7 +30,7 @@
 	<!-- /wp:group -->
 	<!-- wp:group {"align":"wide","layout":{"type":"default"}} -->
 	<div class="wp-block-group alignwide">
-		<!-- wp:pattern {"slug":"twentytwentyfive/news-blog-query-loop"} /-->
```
juanfra, line 33: This template also needs the renaming of the slug.

juanfra (commented):
Thanks, Joen! Works well for me. I found a few tiny things I'm going to push changes for.

patterns/media-instagram-grid.php
```diff
@@ -1,8 +1,8 @@
 <?php
 /**
  * Title: Instagram grid
- * Slug: twentytwentyfive/instagram-grid
- * Categories: media, featured
+ * Slug: twentytwentyfive/media-instagram-grid
+ * Categories: media, gallery, featured
```
carolinan, line 5: ⭐

patterns/services-3-col.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
- * Title: Services - Three Columns
- * Slug: twentytwentyfive/services-three-columns
+ * Title: 3 columns
+ * Slug: twentytwentyfive/services-3-col
```
carolinan, line 4: Is this meant to be Services, 3 columns, like the Pricing, 3 columns?

patterns/template-archive-vertical-header-blog.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * Title: Archive for the right aligned blog
```
carolinan: I think all the right-aligned template names could use some adjusting.
Is it right-aligned, right aligned, or, is it the vertical header that should be in focus?

patterns/services-3-col.php
jasmussen: Yes. One moment.

patterns/template-archive-vertical-header-blog.php
jasmussen: I'll take a stab.

jasmussen:
Pushed some "aligned" improvements, and added "Services, " as a prefix to the 3 columns.

PR #447, opened 2024-09-28
Try: Move the CSS for the outline button :hover style to theme.json

### LINKED ISSUE #413: Remove custom outline button styles
We should use theme.json to style variations instead. Otherwise it complicates how styles are accounted for. 

```
.wp-block-button.is-style-outline
	> .wp-block-button__link:not(.has-text-color, .has-background):hover {
	background-color: var(--wp--preset--color--contrast, transparent);
	color: var(--wp--preset--color--base);
	border-color: var(--wp--preset--color--contrast, currentColor);
}

```

carolinan:
**Description**
In this PR I am moving the CSS for the hover style for the outline button variation to theme.json.
Note, I was not able to make it work with the `>` combinator so I removed it. Open for alternatives.

The PR also deletes the CSS file and the PHP function `twentytwentyfive_block_stylesheets()`

Closes https://github.com/WordPress/twentytwentyfive/issues/413

**Testing Instructions**

Add two buttons with the outline style variation, one with and one without a background color.
Add links to the buttons.
Hover over the button that does not have a background. Expected: The background color changes slightly.
Hover over the button that has a background color. Expected: The background does not change.

### DIFF
--- a/assets/css/button-outline.css
+++ b/assets/css/button-outline.css
@@ -1,3 +1,0 @@
-.wp-block-button.is-style-outline > .wp-block-button__link:not(.has-background):hover {
-	background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 5%, transparent);
-}
--- a/functions.php
+++ b/functions.php
@@ -67,37 +67,6 @@ endif;
 add_action( 'wp_enqueue_scripts', 'twentytwentyfive_enqueue_styles' );
 
 /**
- * Enqueue custom block stylesheets.
- */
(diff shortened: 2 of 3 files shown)


juanfra (approved):
Thanks Carolina, it is working well to me 🚀 

https://github.com/user-attachments/assets/186dc72d-668a-4f50-81f3-9ac089acc0b4

I believe we don't need the combinator to add such specificity. It should work well with these changes, as the CSS is specific enough.

PR #451, opened 2024-09-28
Add a 404 template to the right-aligned blog set

### LINKED ISSUE #443: Add a 404 template with a vertical header

After assigning the design with the vertical header to my templates, I am left with a 404 template that does not have a vertical header.
Can we still add a 404 template that matches the rest of the set?

carolinan:
**Description**

This PR creates a new pattern which uses the vertical header template part and the default hidden-404 in the content section.
The purpose is to make it easier create a site where the header is the same on standard templates.

Closes https://github.com/WordPress/twentytwentyfive/issues/443

**Screenshots**

Selection:
( No idea why the footer is not showing in the preview)
image

Front:
67 local_jkyhkykuykh

**Testing Instructions**
Go to Editor > Appearance > Templates
Select "Page: 404"
Open the settings sidebar. Select the Templates tab.
Open the Design panel.
Select "Right-aligned blog, 404"
Save
View the 404 on the front of the site.
Confirm that the vertical header is used.
Test that the design looks good on different browser widths.

### DIFF
--- a/patterns/template-404-vertical-header-blog.php
+++ b/patterns/template-404-vertical-header-blog.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * Title: Right-aligned blog, 404
+ * Slug: twentytwentyfive/template-404-vertical-header-blog
+ * Template Types: 404
+ * Viewport width: 1400
+ *
+ * @package WordPress
+ * @subpackage Twenty_Twenty_Five
+ * @since Twenty Twenty-Five 1.0
+ */
+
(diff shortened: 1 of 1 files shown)


juanfra (approved):
Thank you, works well. And it make sense to provide a vertical-header of the 404 like we do for the other templates (archive, search, blog, etc).

https://github.com/user-attachments/assets/2d617362-30a2-43fc-bb3b-85dd07fb533d

PR #456, opened 2024-09-28
Fix the block validation error in the link pattern

carolinan:
**Description**
There is a missing closing `</div>` in the pattern for the link post format.
This PR adds it, to resolve the block validation error

**Testing Instructions**

Place a link post format pattern
Check that there is no block validation error


juanfra (commented):
Thank you 😄

Issue #423, opened 2024-09-25
Inconsistencies between parent, color and type style variations


richtabor:
Now that we've been tightening up type (https://github.com/WordPress/twentytwentyfive/pull/422) and colors (https://github.com/WordPress/twentytwentyfive/issues/415) there are many inconsistencies between the color and typography variations, alongside the full style variations. 

I propose we tighten up the color and typography variations individually (removing any values that don't need overriding, fixed typography values—they should use presets like any other style variation—etc), then compose a few for the full style variations. 

Each style variation should be a combination of typography presets and color presets. Otherwise it's very difficult to keep a high bar for these. If we can't do that, then we should only move forward with individual typography and color variations.

richtabor:
### Typography 

Noted that the typography style variations are quite different from the typography values in the full style variations. The same applies for any type style in a full style variation. Those should be available in the typography variations. 

If font sizes change, then the font size presets should be included in the typography variation (just like how color palettes are included in color variations).

richtabor:
If section styles are adapted, like they are in some theme style variations, those same color declarations should be made in the /styles/colors/* style variations. 

If colors are applied to specific blocks, then those should also exist in these color variations.

richtabor:
It may be best to start just with the individual typography and color variations.

carolinan:
The final variation is merged so this work can continue 👍

richtabor:
> The final variation is merged so this work can continue 👍

What do you mean?

richtabor:
For instance, if a full theme style variation is defining a different set of values for section styles, those should be in the color variations as well. As-is, it's going to be very challenging fixings things with so many variables. 

![CleanShot 2024-09-26 at 11 51 12](https://github.com/user-attachments/assets/b6b1445d-2f8c-4cb6-988b-d94b0f191f78)

carolinan:
> > The final variation is merged so this work can continue 👍
> 
> What do you mean?

When this issue was created, all full / combined / theme style variations were not in trunk. Now they are.

carolinan:
The inconsistencies should all be solved (according to the diffs between the files).
There may still be some work left regarding removing duplicate styles, especially as there may be need to be some changes for the search block across all variations.

PR #457, opened 2024-09-28
Try to match noon and typography preset-1

carolinan:
**Description**

## Typography

I copied the font size **setting** from Noon to preset 1.
I copied the body font size **style** from Noon to preset 1.
I copied the block typography **styles** from Noon to preset 1
I copied the element typography **styles** from Noon to preset 1

I changed the format in Noon to use "pipes":  
```
- "fontSize": "var(--wp--preset--font-size--medium)",
+ "fontSize": "var:preset|font-size|medium",
```

**Testing Instructions**
Compare the typography settings and styles in the two files, they should match.
Look for any typography settings or styles that are duplicates, that are the same as in theme.json.

## Color

I copied the entire styles section from the combined 02-Noon to the color preset 02-Noon, and from there, removed anything that was not color related.

### Buttons
I removed the hover styles and added the color-mix to two sections.


carolinan:
I can see why the heading sizes needs to be one size larger than the defaults from theme.json.
I don't have a strong opinion, but the jump in size from h1 to h2 is rather large?

H1 is 58px, (`clamp(2.625rem, 2.625rem +((1vw - 0.2rem)* 1.569), 3.625rem);`)
H2 is 32 (`clamp(1.75rem, 1.75rem +((1vw - 0.2rem)* 0.392), 2rem);`)
image

Compared with the sizes in Figma:
H1 is 42px
H2 is 28.8px
image

carolinan:
This does not look good **in my opinion**. This is from the default single post template:
image

"Written by" is a paragraph and uses Literata, inherited from the body.
The post term uses Beiruti, because in the original Figma design, Beiruti is used for the post **tags** that are below the post
content.

In Figma, the whole line uses Beiruti, font size 20px; https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=77-584&t=U5e7KtwXbVErNU0v-4

carolinan:
On the default home template, the post title is set to X-large in the markup.
Is it too small?

image

carolinan:
There are two ways to sort the section styles from light to dark:
- Update the order of the colors in the palette
- Override the colors in each section style, under styles > variations

Do we want to do either of these?

jasmussen:
This works well for me:

changes

PR #459, opened 2024-09-28
Try to match Afternoon and typography preset 3

carolinan:
**Description**

## Typography

I copied the font size **settings** from Afternoon to preset 3.
I copied the block typography styles from Afternoon to preset 3.
I copied the element typography styles from Afternoon to preset 3.

Testing Instructions
Compare the typography settings and styles in the two files, they should match.
Look for any typography settings or styles that are duplicates, that are the same as in theme.json.

## Color

Accent 1 and Accent 2 did not match Figma; accent 1 was assigned to accent 2 and vice versa. I updated this.
One section style that was in Figma was not in the theme, so I replaced a section style that was duplicate.


carolinan:
In my opinion, it does not look good that the paragraph "Written by" is small and the author is medium:
image

PR #461, opened 2024-09-29
Try to match Twilight and typography preset 4

carolinan:
**Description**

There were many differences between the typography preset and the combined variation. 
I updated some values to match Figma.
I removed the hover styles from the buttons, so that it falls back to the color-mix in theme.json.
The exception is the one button in section 2 that has a different background and text color.
Finally, I increased the padding on the buttons to match Figma, but used rem instead of px.

Testing Instructions
Compare the typography settings and styles in the combined variation and the typography preset, they should match.
Compare the color settings and styles in the combined variation and the color preset, they should match.
Look for any settings or styles that are duplicates, that are the same as in theme.json.


carolinan:
Just like in some other variations, I don't think the "written by" in the default single post template looks good: It is inconsistent in size, font family and capitalization:
image

PR #476, opened 2024-09-30
Fix #465: Tags overlap when extending to two lines

kartikmehta8:
**Description**
When the number of tags extend to two lines or more, the pills overlap with the row above/below.

This PR will fix this tags overlapping giving appropriate line height.

**Screenshots**

Screenshot 2024-09-30 at 16 40 24

**Testing Instructions**

1. Create a new post.
2. Add Title, content and tags.
3. Visit the post and see it on smaller screens.

Closes #465

PR #478, opened 2024-09-30
Fix #466: Multi-word tags break to multiple lines

### LINKED ISSUE #466: Multi-word tags break to multiple lines
**Description**

When the tags extend to multiple lines, there is a possibility that the tag at the end of the first row is split between two lines if it is a multi-word tag.

Because the tag is styled like a pill, it appears broken. 

1. Go to https://2025.wordpress.net/2024/09/06/vash-gon-jicarilla-by-edward-s-curtis/ on a mobile device. 
3. Scroll down to the tags. 

**Expected behavior**

If the tag is styled as a pill, the pill should remain connected. 

**Screenshots**

image

**Environment info**

 - Device: iPhone 16 Pro
 - OS: iOS 18
 - Browser: Safari

kartikmehta8:
**Description**
When the tags extend to multiple lines, there is a possibility that the tag at the end of the first row is split between two lines if it is a multi-word tag.

Because the tag is styled like a pill, it appears broken.

This PR will add a property of `no-wrap` for the link tags (pills) that will not break on multiple lines on every screen size.

**Screenshots**
Screenshot 2024-09-30 at 17 08 47

**Testing Instructions**

1. Create a post.
2. Add title, content and multi word tags.
3. Visit the post on various screens.

Closes #466

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -1154,7 +1154,7 @@
 				}
 			},
 			"core/post-terms": {
-				"css": ".wp-block-post-terms a { white-space: nowrap; }",
+				"css": "& a { white-space: nowrap; }",
 				"typography": {
 					"fontSize": "var:preset|font-size|small",
 					"fontWeight": "600"
--- a/theme.json
+++ b/theme.json
@@ -1154,6 +1154,7 @@
 				}
 			},
 			"core/post-terms": {
+				"css": ".wp-block-post-terms a { white-space: nowrap; }",
 				"typography": {
 					"fontSize": "var:preset|font-size|small",
 					"fontWeight": "600"


carolinan:
Thank you for working on this. This is working well, but I would like to await input from others on whether this should be applied to all terms blocks or only the pill style.

This block does not have a wrap/no wrap option in the block settings sidebar like some other blocks do, so using the custom CSS is the only option. If we need to, we can also discuss if it should be in the JSON file, or in a CSS file.

If it is used for the terms block, in theme.json, the CSS will show up in the global styles sidebar and the user will be able to remove it:
image

carolinan:
I'd also want to know if the duplicate block class name is necessary:

```
:root :where(.wp-block-post-terms.wp-block-post-terms a ) {
    white-space: nowrap;
}
```

theme.json
```diff
@@ -1154,6 +1154,7 @@
 				}
 			},
 			"core/post-terms": {
+				"css": ".wp-block-post-terms a { white-space: nowrap; }",
```
juanfra: Let's use the ampersand to avoid the duplicate selector.

```suggestion
				"css": "& a { white-space: nowrap; }",
```

juanfra (commented):
Looking good! Thank you for working on this. I left a suggestion.

I'm wondering if we should also use `display: inline-block` for these, so that they don't look on top of each other when viewed on mobile. @carolinan @beafialho thoughts?

carolinan:
> Looking good! Thank you for working on this. I left a suggestion.
> 
> I'm wondering if we should also use `display: inline-block` for these, so that they don't look on top of each other when viewed on mobile. @carolinan @beafialho thoughts?

Isn't that managed by the other issue? https://github.com/WordPress/twentytwentyfive/pull/476

juanfra:
> Isn't that managed by the other issue?

@carolinan looks like it. I saw it after I left the comment. This one should be good to merge.

Issue #495, opened 2024-10-03
Link underline inconcistencies


carolinan:
**Description**

In the default single post template on Figma, the "Written" by text uses two different colors: `#686868` for text and `#111111` for the links. Both the author name and the category name are links that are underlined.

![Image](https://github.com/user-attachments/assets/9dfb33dc-3ef0-4258-a5b1-029edf2f6b44)

https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3326-1728&t=GUAp7xV6mAMGjrcI-4

In the theme, well, the text is so small that I can't see the color differences between the text and the link without zooming in;
but I can tell that the category is not underlined.
For the category to match the author name, the link should be underlined, and the underline should not be removed on :hover.
Zoom in 200%:
![Image](https://github.com/user-attachments/assets/7db33288-4d44-4b05-95ef-013d48a08a1a)

carolinan:
Could the "in" here be removed? It is making the string difficult to translate because it has four separate parts with the block markup included.
Perhaps the middle dot separator could be used between the username and the category?

carolinan:
In the photo blog single post template, the author name is[ not underlined in Figma,](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3378-2643&t=7X9PGm4sFRbbYZPb-4)
![Image](https://github.com/user-attachments/assets/832a1c86-d027-4cb4-acfe-e205c128d071)

But it is underlined in the theme:

![Image](https://github.com/user-attachments/assets/7ac46b2b-fc0b-46c5-874c-aefb9045f812)

Update: It also looks like the font weight does not match. It is 400 in Figma and 300 in the theme.

carolinan:
In the single post template for the news blog, the author is not underlined: Because it is not linked to the author archive. I think this needs to be more consistent and that the link should be added.

carolinan:
Single offset post template: In Figma, the post date is underlined. But In the theme, the post date is not linked.
None of the other single post templates has a post date that links to the date archive. @beafialho should this be added or not?

PR #496, opened 2024-10-03
Multiple templates: Fix underline inconsistencies

carolinan:
**Description**
Partial for: https://github.com/WordPress/twentytwentyfive/issues/495

## Default single post template
- Enables the underline typography setting on the terms block, in the markup of the template.
- Moves the color definition from the two paragraphs to the wrapping group block.

**Screenshots**
After: Both links are underlined by default.
image

**Testing Instructions**
View the default single post template and confirm if both the author name and category are underlined by default.
Confirm if the heme matches [the design in Figma.](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3326-1728&t=GUAp7xV6mAMGjrcI-4)

## Photo blog single post template
By changing the default text decoration on the post author name block to "none", in theme.json, the theme matches Figma.

**Screenshots**
After: the link is not underlined by default.
image
⚠️  **NOTE: This matches the design, but the link is invisible to sighted users. It is not accessible. The link can only by found by users who will make assumptions and recognize the pattern of the author name and category being linked.

**Testing Instructions**
View the photo blog single post template and confirm that the author name is not underlined by default.

## News blog single post template
The author name and avatar were not linked. The links were added to make the template more consistent with the other templates.

⚠️  To match Figma, no underline was added.  A sighted user will only find the links if they recognize the pattern of the author name being linked (and goes to try it).

image

**Testing Instructions**
View the News blog single post template and confirm that the author name and avatar are linked to the author's user profile.

## Right-aligned blog with vertical header, single post template
The author name and avatar were not linked. The links were added to make the template more consistent with the other templates.

⚠️  To match Figma, no underline was added.  A sighted user will only find the links if they recognize the pattern of the author name being linked (and goes to try it).

image

**Testing Instructions**
View the single post template for the right-aligned blog, and confirm that the author name and avatar are linked to the author's user profile.

### DIFF
--- a/patterns/hidden-written-by.php
+++ b/patterns/hidden-written-by.php
@@ -15,7 +15,7 @@
 	<!-- wp:paragraph -->
 	<p>Written by </p>
 	<!-- /wp:paragraph -->
-	<!-- wp:post-author-name {"isLink":true} /-->
+	<!-- wp:post-author-name {"isLink":true,"style":{"typography":{"textDecoration":"underline"}}} /-->
 	<!-- wp:paragraph -->
 	<p>in</p>
 	<!-- /wp:paragraph -->
--- a/patterns/hidden-written-by.php
+++ b/patterns/hidden-written-by.php
@@ -10,15 +10,15 @@
  */
 
 ?>
-<!-- wp:group {"style":{"spacing":{"blockGap":"0.2em","margin":{"bottom":"var:preset|spacing|60"}}},"fontSize":"small","layout":{"type":"flex","flexWrap":"nowrap"}} -->
-<div class="wp-block-group has-small-font-size" style="margin-bottom:var(--wp--preset--spacing--60)">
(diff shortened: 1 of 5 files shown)


carolinan:
I already added this comment on the issue, but perhaps it is more visible here:

Could the "in" in the "Written by..." on the default single post be removed? It is making the string difficult to translate because it has four separate parts with the block markup included.
Perhaps the middle dot separator could be used between the username and the category?

beafialho:
Thank you for working on the inconsistencies @carolinan. The "in" should be kept, ideally.

**Default single post template**: looks good ✅
**Photo blog single post template**: looks good ✅

The other two, could you share a screenshot of what it's looking like for you? I'm not sure I'm seeing the same.

carolinan:
Yes, of course. I will add them to the PR description.

beafialho:
Thank you! The other two look good as well.

Issue #499, opened 2024-10-04
Pricing, 3 columns: Remove invalid rel from button blocks


carolinan:
**Description**
There seems to have been a misunderstanding about how to add button blocks to patterns.

Two button blocks in the "Pricing, 3 columns" pattern have the `rel` atribute set to a hashtag which is not valid. One has a rel attribute which is empty.
`Rel` shouldn't be needed at all since these are placeholder buttons.

PR #502, opened 2024-10-04
fix: removed rel from buttons in 3 col pricing pattern

troychaplin:
**Description**

To fix issue #499 and remove rel tags from buttons in a pattern

**Testing Instructions**

1. Create new page
2. Add Pricing 3 Column pattern
3. Inspect buttons, or switch to editor mode
4. Rel should not exist on buttons


carolinan:
Thank you

PR #507, opened 2024-10-07
Evening, section style 3: update link color to maintain contrast ratio

carolinan:
**Description**
While testing, I found that the link color in section style 3 in the "Evening" variation had too low contrast ratio against the background.

This PR updates the link color in this section to be the same as the text color, using the `contrast `color.
The resulting contrast ratio is 4.61:1 which passes WCAG level AA.

image

**Testing Instructions**

In the Site Editor, select the evening Style variation.
Insert a group block and apply section style 3.
In the group, insert a paragraph with a link.

Confirm that the link color is readable.
The contrast ratio should be at or above 4.5:1.

### DIFF
--- a/styles/01-evening.json
+++ b/styles/01-evening.json
@@ -131,6 +131,15 @@
 					}
 				}
 			},
+			"section-3": {
+				"elements": {
+					"link": {
+						"color": {
+							"text": "var:preset|color|contrast"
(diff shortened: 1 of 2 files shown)


carolinan:
The visible change here is minimal, so I am merging this.

PR #510, opened 2024-10-07
Fix/banner book desc heading

troychaplin:
**Description**

Pattern uses an H3, creates incorrect heading level a11y error

**Testing Instructions**

1. Add "Banner with book description" pattern to a page
2. Check H2 is used

### DIFF
--- a/patterns/banner-about-book.php
+++ b/patterns/banner-about-book.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Title: Banner with book description
  * Slug: twentytwentyfive/banner-about-book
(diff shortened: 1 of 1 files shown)


patterns/banner-about-book.php
```diff
@@ -18,8 +18,8 @@
 	<div class="wp-block-columns alignwide">
 		<!-- wp:column {"verticalAlignment":"center","width":""} -->
 		<div class="wp-block-column is-vertically-aligned-center">
-			<!-- wp:heading {"level":3,"className":"wp-block-heading","fontSize":"xx-large"} -->
-			<h3 class="wp-block-heading has-xx-large-font-size"><?php esc_html_e( 'About the book', 'twentytwentyfive' ); ?></h3>
+			<!-- wp:heading {"level":2,"className":"wp-block-heading","fontSize":"xx-large"} -->
```
carolinan: Nit: H2 is the default for the heading block, so the level is not needed.
```suggestion
			<!-- wp:heading {"className":"wp-block-heading","fontSize":"xx-large"} -->
```

Issue #403, opened 2024-09-25
Add metadata: Name sections of blocks


carolinan:
**Is your feature request related to a problem? Please describe.**
In https://github.com/WordPress/twentytwentyfive/issues/402, I am giving an example of why the pattern metadata should be removed.

In this issue, I would like to propose that metadata, specifically block names, can be added **deliberately** instead, to make it easy for users to manage sections in the ListView.

richtabor:
> In this issue, I would like to propose that metadata, specifically block names, can be added deliberately instead, to make it easy for users to manage sections in the ListView.

100%

A consistent convention could just be to use the category as the top-level block name. I.e. "Call to action" instead of "CTA: Grid layout with products and link"

unscripted:
@carolinan I'm pretty sure I follow, but for clarity, would you mind providing a few examples of what it is now vs what it would become?

carolinan:
@unscripted Blocks can be named by adding the name meta data to the comment with the JSON attributes:
`"metadata":{"name":"Example"}`

```
<!-- wp:group {"tagName":"main","metadata":{"name":"Main"},"style":{"spacing":{"margin":{"top":"var:preset|spacing|60"}}},"layout":{"type":"constrained"}} -->
<main class="wp-block-group" style="margin-top:var(--wp--preset--spacing--60)">
```

The name shows in the ListView and in the block settings sidebar:
![image](https://github.com/user-attachments/assets/5e66dbb0-8ed1-4ed7-b783-8aede63afe34)

carolinan:
I am blocking this temporally while researching if the name metadata is translatable.

If it is not already translatable by the parsing of the name, then
- The name can not be added in the HTML files.
- The name needs to be wrapped inside translation functions in the PHP files:

`"metadata":{"name":"<?php echo esc_html_x( 'Main', 'HTML tag name', 'twentytwentyfive' ); ?>"},`

carolinan:
I am going to remove the block, but recommend that the theme does not add names to the sections.
If anyone wants to test the PHP approach above, you are welcome to, I would love to hear the results.

Link to Slack discussion: 
https://wordpress.slack.com/archives/C02QB2JS7/p1727845190074579

>The PHP approach could work, but needs testing to verify it doesn‘t break when changing locales etc.

richtabor:
> I am blocking this temporally while researching if the name metadata is translatable.

> If it is not already translatable by the parsing of the name, then

[...]

PR #514, opened 2024-10-08
Full page patterns: Improve zoom out

### LINKED ISSUE #497: Zoom out is not working on full page patterns
**Description**

I have been testing TT5 with zoom out and I have encountered a problem with full page patterns:

Image

Because there is a group containing all the patterns, zoom out detects that as a full section, and it doesn't let you insert patterns in between. If we remove that wrapper group:

Image

The problem is that if we remove that wrapper group, we miss the `blockGap` value.

I can't suggest a solution for this, there are options:

- We can remove the wrapper from the full page pattern and solve the `blockGap` issue on a per pattern basis. This would be the easy solution. Probably only the patterns with colored backgrounds need their margins set to 0 for this to work
- The problem with the above solution is that it's not universal, so other themes doing these face the same issue. The problem is that there is no way for zoom out to know that it should ignore the wrapper, unless we set the tag to `main`, which I think it's also an option here, since it makes sense semantically, but I'll let the accessibility experts weigh in on that

/cc @richtabor

carolinan:
**Description**

This PR updates the block patterns that are designed as full page layouts.  These patterns were built using one group that wrapped all inner patterns. The PR removes the outer group and instead places each pattern inside its own group.
These new groups have top and bottom margin set to 0, to remove the blank space between the sections.

This ensures that each inner pattern show as a separate section when zoom out is enabled, and the spacing is maintained.

Closes https://github.com/WordPress/twentytwentyfive/issues/497

**Screenshots**

https://github.com/user-attachments/assets/8454f06a-7a8e-4647-b62a-b83679c8406c

**Testing Instructions**

Create a new page.
In the pattern selection modal, select a pattern that has multiple sections, for example the business homepage.
Confirm that visually, there is no change before and after the PR.
Select and add background colors to all patterns: Confirm that there is no vertical space between the colored sections.
Enable zoom out. Confirm that you can add patterns between the sections.
- Repeat for all patterns updated by this PR.

### DIFF
--- a/patterns/page-business-home.php
+++ b/patterns/page-business-home.php
@@ -15,13 +15,39 @@
  */
 
 ?>
-<!-- wp:group {"align":"full","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
-<div class="wp-block-group alignfull">
+
+<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0">
 	<!-- wp:pattern {"slug":"twentytwentyfive/cta-centered-heading"} /-->
+</div>
+<!-- /wp:group -->
+
+<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0">
 	<!-- wp:pattern {"slug":"twentytwentyfive/overlapped-images"} /-->
+</div>
+<!-- /wp:group -->
+
+<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0">
 	<!-- wp:pattern {"slug":"twentytwentyfive/services-3-col"} /-->
+</div>
+<!-- /wp:group -->
+
+<!-- wp:group {"align":"full","style":{"spacing":{"margin":{"top":"0","bottom":"0"}}},"layout":{"type":"constrained"}} -->
+<div class="wp-block-group alignfull" style="margin-top:0;margin-bottom:0">
 	<!-- wp:pattern {"slug":"twentytwentyfive/testimonials-large"} /-->
+</div>
+<!-- /wp:group -->
+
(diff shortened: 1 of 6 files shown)


carolinan:
I was not able to get a recording of this where the patterns don't show outside the zoom out area. Deactivating zoom and and enabling it again still shows the pattern outside the area, but by a smaller amount.

jasmussen:
Tested the outlined patterns, this is working well for me. Here's just one example:

test

Separate from this PR, we probably need to tweak the Stories pattern, this one:

Screenshot 2024-10-08 at 09 43 48

It seems to break onto two lines in very many cases, unfortunately, so we might need to reduce the font size a bit.

juanfra (commented):
Thanks Carolina, it's working well for me. I was able to use the zoom out mode with these landing pages, the gap is not there and I'm able to insert patterns in between.

One thing I noticed, probably unrelated to this PR and a problem with zoom out is that when the list view or inserter sidebar are not there, the zoom out has some sort of left/right padding. I'm on the latest Gutenberg (trunk)

https://github.com/user-attachments/assets/24b26d0a-d15f-49ae-b0e1-8101fb37ef55

jasmussen:
@getdave would you be able to briefly take a look at juanfra's comment above, see if this is a known issue in the block editor or not?

richtabor:
It's not 100% ideal, but it works. Can we add names to these groups though, so it's more helpful to scan and identify patterns?

CleanShot 2024-10-08 at 23 57 35

carolinan:
@richtabor The problem with the names is that it was discovered that they are not translatable.

getdave:
> @richtabor The problem with the names is that it was discovered that they are not translatable.

Oh that's not good. 

The names are stored in the `metadata.name` block attribute. Is there an Issue tracking that problem?

getdave:
Oh hold on. These are `.php` files. So cant we just use PHP to add the `attributes.metadata.name` and run it through a translation function at point of output?

```php
<!-- wp:group {"metadata":{"name":"<?php _e( 'My custom name here', 'my-text-domain' ) ?>"},"align":"full","className":"is-style-section-5","style":{"spacing":{"padding":
```

carolinan:
It needs to be tested. Please see the linked conversation here: https://github.com/WordPress/twentytwentyfive/issues/403

carolinan:
There are already attributes in other pattern files that have the PHP functions added inside, like the aria labels for the search block.

richtabor:
> It needs to be tested. Please see the linked conversation here: https://github.com/WordPress/twentytwentyfive/issues/403

Left a comment there. We should really consider adding names to these—they're much more identifiable in list view with names.

qasumitbagthariya:
Upon thoroughly reviewing on my end, everything appears to be in order. The zoom functionality is working as expected, including proper interaction with the group name. 

https://github.com/user-attachments/assets/1dab03b6-dd01-4d55-93d6-04e723086f02

annezazu:
Thank you for handling this 🙌🏼

PR #516, opened 2024-10-08
Fix: More link has incorrect alignment

### LINKED ISSUE #491: More link in posts list has incorrect alignment.
**Description**

In the posts list, the '(more...)' link aligns to the left of the content area, rather than the left of the block area. This is because it's a direct descendant of `.is-layout-constrained` but is not a block level element. The `max-width` and `margin` attributes applied in `.is-layout-constrained > :where(:not(.alignleft):not(.alignright):not(.alignfull))` don't apply because the link is an inline element.

1. Activate the theme.
2. Set the front page to be a posts page.
3. Add a post that is long enough to trigger a more link.

**Expected behavior**

The link should align to the left edge of the post excerpt.

**Screenshots**

Image

**Environment info**

Please share the browser(s) are you seeing the problem on.
Please share the Device you are using and operating system (e.g. "Desktop with Windows 10", "iPhone with iOS 14", etc.).
 - Device: Windows laptop
 - OS: Windows 11
 - Browser: Chrome 
 - Version: 129.0.6668.70 (Official Build) (64-bit)

carolinan:
**Description**

This PR adds CSS to style.css to turn the more link (the more block) into a block level element.
The purpose is to ensure that:
- By default, the block is contained within the default content width, aligning with the rest of the post content.

Closes https://github.com/WordPress/twentytwentyfive/issues/491

**Screenshots**
Before, the more block was all the way to the left:
Blog home page showing a post with the read more block all the way to the left

After, the more block is aligned with the rest of the content
Blog home page showing a post with the read more block contained within the default width.

**Testing Instructions**
Reset any changes you may have made to the blog home template.
Create a new post.
Insert a read more block.
View the post you created in the list of posts, on the blog home template.
Confirm that the block is aligned with the other post content.

### DIFF
--- a/style.css
+++ b/style.css
@@ -36,3 +36,11 @@ a {
 h1, h2, h3, h4, h5, h6, blockquote, caption, figcaption, p {
 	text-wrap: pretty;
 }
+
+/*
+ * Change the position of the more block on the front, by making it a block level element.
+ * https://github.com/WordPress/gutenberg/issues/65934
+*/
+.more-link {
+	display: block;
+}


juanfra (approved):
Thank you! Looks good to me

https://github.com/user-attachments/assets/e6041c70-e1e9-42da-8c5e-baba8e8a72a8

Issue #501, opened 2024-10-04
Quote & Pullquote visual adjustments


beafialho:
We've come down to the finest details. While testing the theme, I noticed the quote isn't looking as it was before. The font size is smaller than the rest of the text, and it should be the same size. It's also using Semi Bold font weight, this was probably intentional, but if possible, let's make the change to use the same weight as the rest of the text as well.

| Currently      | Improved |
| ----------- | ----------- |
|![Image](https://github.com/user-attachments/assets/ffc4502d-56e5-41e6-9644-9e57ae2bed38)    |    ![Image](https://github.com/user-attachments/assets/55262c3f-7c4e-4869-95dc-e8ab65015acf)  |

Citation is also looking different in quote and pull quote. They should look similar in both. Notice the differences in the two citations: 

![Image](https://github.com/user-attachments/assets/fa2a5c37-3b6a-4977-9c45-551fa3eb53c5)

This bit that's applied to the pull quote seems good to be applied to both. We should however apply `-0.1px` letter spacing to both.
```
"elements": {
	"cite": {
		"typography": {
			"fontSize": "var:preset|font-size|small",
			"fontStyle": "normal"
				}
```

troychaplin:
@beafialho is the idea here that the citations would look the same in both? IMO it certainly looks better than the pullquote citation being a large font with a matching -0.1. Adding a couple attachments

![Image](https://github.com/user-attachments/assets/fb633003-5442-49e2-b318-c01bf0944414)

![Image](https://github.com/user-attachments/assets/766e6587-8c40-4992-9dc8-5eab3403c30f)

troychaplin:
I have a branch in my fork ready to PR in depending on which style is needed here. Branch currently setup with have citations matched up on both blocks -- https://github.com/troychaplin/twentytwentyfive/tree/update/quote-pullquote-visuals

beafialho:
Yes, citation should look the same size and letter spacing in both, not huge like the quote.

beafialho:
Closing as this has been addressed.

PR #517, opened 2024-10-08
Theme.json visual adjustments for quote, pullquote, cite & code block

beafialho:
This PR addresses #501 and adds an enhancement to the code. This is the intended look of the blocks worked on:

- Improved letter spacing, font size and font weights

Captura de ecrã 2024-10-08, às 09 40 27

Captura de ecrã 2024-10-08, às 09 40 31

Captura de ecrã 2024-10-08, às 09 41 25

### DIFF
--- a/styles/01-evening.json
+++ b/styles/01-evening.json
@@ -67,14 +67,6 @@
 					}
 				}
 			},
-			"core/code": {
-				"typography": {
-					"fontSize": "var:preset|font-size|medium",
-					"fontStyle": "normal",
-					"fontWeight": "300",
-					"letterSpacing": "0px"
-				}
-			},
 			"core/site-tagline": {
 				"color": {
 					"text": "var:preset|color|accent-4"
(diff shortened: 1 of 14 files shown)


carolinan:
@beafialho All other presets and the combined variations are unchanged, is that correct?

carolinan:
I am seeing some duplicates, there are presets that set the cite to font size small and font weight 300, these can be removed since they now match theme.json.

beafialho:
Correct @carolinan, I only changed the default one. However, now that I tried it, the Evening variation should have the same the same settings, since the only thing varying is its colors.

carolinan:
I think I have found and removed all duplicate styles.

carolinan:
@juanfra Aside, I think we can remove all `"fontStyle": "normal"` since this is the default.
The only place where it is needed is in theme.json settings > typography > fontfamilies

beafialho:
Thank you for the help! Looking good 🚀

juanfra (approved):
Thank you, looks good to me 🫡

PR #522, opened 2024-10-08
Fix: Color problems with "CTA with grid layout with products and link"

### LINKED ISSUE #470: Color problems with "Call to action with grid layout with products and link"
**Description**

In some of the style variations, the text color inside the cover block is not visible enough.

Select the "Twilight" style variation.
Insert a "Call to action with grid layout with products and link" pattern.
Confirm that the text between the images has a color contrast that is too low against the background:

image

Repeat the same for "Evening":
image

And for Sunrise:
image

In Twilight, it is kind of visible but does not live up to the text-to-background color contrast requirement. The ratio is  2.26:1.
image

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/470

Update cover blocks to use contrast color for texts to work in different variations.

**Screenshots**

https://github.com/user-attachments/assets/25040d98-233e-435f-b8f5-c40ccbd10136

**Testing Instructions**

1. Create a page.
2. Insert the "Call to action with grid layout with products and link" pattern
3. Test the page with the different style variations.
4. Confirm that it looks good and the colors follow a11y guidelines.


carolinan (approved):
I have tested the pattern with all combined style variations and the text is readable now, thank you.

Issue #485, opened 2024-10-02
Inconsistent search block styles in the style variations


carolinan:
**Description**
There are some differences in styling of the button blocks and the search form when the different style variations are applied.

According to [the design in Figma](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=77-7923&m=dev), these variations should have input elements and buttons with a border radius that is not the same as the default:

- Dusk: 4px
- Afternoon: 0
- Morning: 0
- Sunrise: 0
- Midnight: 0

**Step-by-step reproduction instructions**
Add button block.
Add a couple of search form blocks with different button settings:
Outside, inside, with icon, etc.

**Expected behavior**
For the theme to match the design in Figma.

**Screenshots**
Noon:
Missing shadow (maybe intentional?)
Different font size: the search form button has larger text and a different font family (maybe intentional?)
![image](https://github.com/user-attachments/assets/e3ab9184-9cea-4057-8d9a-33d109c3c599)

Dusk:
Different border radius
![image](https://github.com/user-attachments/assets/da8f1e90-bc5b-4e04-bc70-cdef9ea54182)

Afternoon:
Different border radius
![image](https://github.com/user-attachments/assets/4f141038-38a9-4277-a13f-4abf4b2fa9e3)

Morning:
Different border radius
![image](https://github.com/user-attachments/assets/aa5aa0dd-ba64-4f2b-95a9-5609a9e95238)

Sunrise:
Different border radius
![image](https://github.com/user-attachments/assets/d3f8ced6-e013-4919-b694-41c8dfc36c1e)

Midnight:
Different border radius
![image](https://github.com/user-attachments/assets/f12d58d5-8a92-4596-89bb-619ab8aecff3)

beafialho:
Yes, correct. I believe this is a Gutenberg bug. What other design feedback is helpful here?

carolinan:
I'll try to fix the radius.

Noon:
Is the missing shadow on the search form button intentional?
The search form button has larger text and a different font family, is it intentional?

beafialho:
> Is the missing shadow on the search form button intentional?
> The search form button has larger text and a different font family, is it intentional?

The missing shadow isn't intentional, it should have it. The font family and size should also be the same as the other buttons.

carolinan:
I am not sure about setting the border radius of the inputs to 0 anymore.
[...]

Issue #484, opened 2024-10-02
Discussion: Adding default styles for form elements


carolinan:
**Description**
The Gutenberg PR to add the form inputs to _theme.json elements_ was not completed in time for WordPress 6.7 Beta 1.
Should the theme add default styles for form inputs?

The search and comment forms are already styled.
Login/logout block is not styled:  It can be improved, but needs custom CSS since the expected class names for the button style etc. is missing (this is a known issue).
![image](https://github.com/user-attachments/assets/fe696550-7c55-475b-9df1-d2ef4920239f)

Many forms in third-party plugins and custom blocks are not styled.

mikemcalister:
I've been reading through the previous discussions on this and there are a lot of good points for and against. 

Personally, I'm of the mind that **we should** have some sensible fallback styles for inputs. As Carolina mentioned, we already have some for the comment form, so the styles are there.

As users have pointed out, without any styling the form elements look broken or woefully out of place against the backdrop of the otherwise beautiful design. Maybe that's just the designer in me. 🙃

This should be something we can seamlessly transition to theme.json powered styles when they are available.

justintadlock:
The fix for the login form is just a few lines of CSS: https://github.com/WordPress/gutenberg/issues/50466#issuecomment-1612136221

☝ I've also shared a fix for applying the `.wp-element-button` class to the login form button in that ticket too.

carolinan:
Since it is not a bug introduced in 6.7, I am not sure an update to login/out can be included now, even if someone was able to work on it.

evrpress:
+1 for adding some fallback styles. Things to consider:

**The fallback is for this theme only** 
This would be only for the current theme so others could copy this approach.

**Form plugins often bring their own styles therefore fallback should be minimal with low specificity**
There are not many forms in core but almost every website uses forms. Using a low specificity should prevent any significant issues.

**Comment form styles should be used**
I see no reason why the existing style from the comment form shouldn't be used.

## Discussions

There are already discussions/pr going on
https://github.com/WordPress/gutenberg/issues/29167
https://github.com/WordPress/gutenberg/issues/34198
https://github.com/WordPress/gutenberg/pull/51337

[...]

PR #525, opened 2024-10-09
Search block styles in the style variations, option 1

carolinan:
**Description**

This PR changes the border radius of the search block button and search input field.
It also sets the border color of the search input field to accent color 6, the color with the opacity.
In one variation, it also updates the font family for the button.

With this option, the input field styles are limited to this single block, and managed in multiple theme.json files.

**Screenshots**
Noon, after

Search block styles in the Noon variation

Dusk, after:
Search block styles in the Dusk variation

**Testing Instructions**

Add a search form.
Confirm that the style matches the expectations listed in this issue: https://github.com/WordPress/twentytwentyfive/issues/485
Repeat the test for all style variations


beafialho:
I'm having issues with my testing setup and I don't see the same as your screenshot 😞 However, if that's how it looks for you, it's looking as intended.

carolinan:
Yes the shadow for example is working well for me on the search block button.

But I am concerned about how well the border radius on the input works with other inputs, mainly the comment form which  always has a border radius of 4px.

carolinan:
Like in this screenshot
inputs

beafialho:
> the comment form which always has a border radius of 4px.

Visually, the comment form radius should inherit the styles of each variation.

Also, if we could remove this option for the search block, it would be good. It looks less than ideal in every variation.

Captura de ecrã 2024-10-09, às 11 41 44

carolinan:
There is no way for the theme to remove this option, that I am aware of 🤔 Not even with PHP or JavaScript.

carolinan:
>Visually, the comment form radius should inherit the styles of each variation.

I agree with this. Short term I can try to add it to this PR.

I am still thinking of how we could do this for all inputs including the third party blocks as requested here:
https://github.com/WordPress/twentytwentyfive/issues/484. Yes, I opened this issue, but because it was requested by a user on Twitter.
In case you missed it, especially the select list looks bad in my opinion.

beafialho:
Yes, I saw that. It would be great if we could do it for all inputs.

> In case you missed it, especially the select list looks bad in my opinion.

I have [replied with a visual suggestion](https://github.com/WordPress/twentytwentyfive/issues/484#issuecomment-2401747105) for the select list.

carolinan:
If we handle the comment form and the select list in separate PR's, this is ready for review.

styles/08-midnight.json
```diff
@@ -225,8 +225,13 @@
...
 			"core/search": {
-				"border": {
-					"radius": "0px"
+				"css": "& .wp-block-search__input{border-radius:0px;border-color:var(--wp--preset--color--accent-6);}",
+				"elements": {
+					"button": {
+						"border": {
+							"radius": "0px"
+						}
+					}
```
carolinan, line 234: This is not valid, is it? These are not typography styles.

juanfra (commented):
It looks well. I noticed that when you switch the style variation to one of these and then go back to one where the input has a rounded shape, then that doesn't take effect. For example, trying to go back to the default style.


https://github.com/user-attachments/assets/5bdd6d04-2a6d-4c7a-9f6a-ef445d663933

carolinan:
How strange.
It looks like the variations that are supposed to have 0 border radius on the inputs still have 4px.

My guess would be that it is the styles> blocks > block name > CSS that is not clearing / refreshing.

carolinan:
Because of this bug, I don't think it is worth implementing this change. I think it might be too confusing that the style changes depending on what order you preview the theme variations in...

carolinan:
This still needs a decision.

beafialho:
The default variation looks ok, as well as Noon and Dusk, but the other input fields aren't matching the buttons.

carolinan:
How are you testing it?
Are you saving and resetting the style variation every time you are changing variation?

If you are not saving and resetting, then you are running into the the editor bug that Juan mentioned, and this is why I think this change should not be implemented, the bug is too confusing.

beafialho:
I'm saving the variation, seeing on the front end, I go back to the editor and pick another variation.

carolinan:
Yes, that's it. When we don't click on "reset styles" in between, the CSS for the input field is not updated, that is the bug.
I think this bug is blocking.

What we can do is open an issue for it on Trac, so that once it is fixed in the editor, the theme can be updated.
(There is already an issue for it in the Gutenberg GitHub repository)

PR #528, opened 2024-10-09
Fix: Color problems with the newsletter sign-up pattern

### LINKED ISSUE #526: Color problems with the newsletter sign-up pattern
**Description**

The color contrast ratio in the "Newsletter sign-up" pattern is not high enough in all combined style variations.
The text color on this cover block is set to `#000` which does not work for all accent-color 1 background colors in all palettes.

**I am wondering, if one solution would be to change the order of the palette so that accent-color 1 is always a light color...?**

Insert the "Newsletter sign-up" pattern.
Test the pattern with all combined style variations.

**Expected behavior**
For the text in the pattern to be readable.
WCAG 2.0 level AA requires a contrast ratio of at least 4.5:1 for normal text and 3:1 for large text.

**Note**
Please remember that automated tools are not able to get the correct colors from the cover block, some manual testing is required.

Midnight:
The background and the text color of the heading and text has a contrast ratio of 2.94:1.

Image

Evening:
The background and the text color of the heading and text has a contrast ratio of 3.99:1.
This means that the heading passes, since it has a larger font size, but the paragraph does not pass.

Image

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/526

Update cover blocks to use contrast color for texts to work in different variations.

**Screenshots**

https://github.com/user-attachments/assets/d348bbf1-7b34-448a-af9d-a1eebca959c7

**Testing Instructions**

1. Create a page.
2. Insert the "Newsletter sign-up" pattern
3. Test the page with the different style variations.
4. Confirm that it looks good and the colors follow a11y guidelines.


beafialho:
I'm trying to test this, but I'm having some issues with my local environment...

beafialho:
Is this one looking unaccessible to you as well?

Captura de ecrã 2024-10-09, às 10 37 37

juanfra:
> Is this one looking unaccessible to you as well?

@beafialho yes, that's "Sunrise" right?

carolinan:
The only two color contrasts that are guaranteed to work on all palettes are contrast and base.

beafialho:
> > Is this one looking unaccessible to you as well?
> 
> @beafialho yes, that's "Sunrise" right?

Correct.

carolinan:
From the issue:
> Another important thing to consider is that I believe the color is meant to be the same as in the (stories) footer?

Because the stories footer and the newsletter sign-up has the same background color.

juanfra:
Then, let's consider using the group block with `Style 3` instead of the cover block. I will adjust the pattern for that.

carolinan:
Great point, I am not sure why the cover was used.

juanfra:
@beafialho @carolinan this one is ready for a new pass when you have time. Thanks 🙌

patterns/cta-newsletter.php
```diff
@@ -11,24 +11,28 @@
...
-<aside class="wp-block-cover alignfull is-light" style="padding-top:var(--wp--preset--spacing--50);padding-right:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--50);min-height:460px"><span aria-hidden="true" class="wp-block-cover__background has-accent-1-background-color has-background-dim-100 has-background-dim"></span><div class="wp-block-cover__inner-container">
-	<!-- wp:heading {"textAlign":"center","fontSize":"xx-large"} -->
-	<h2 class="wp-block-heading has-text-align-center has-xx-large-font-size"><?php esc_html_e( 'Sign up to get daily stories', 'twentytwentyfive' ); ?></h2>
-	<!-- /wp:heading -->
+<!-- wp:group {"tagName":"aside","metadata":{"categories":["call-to-action"],"patternName":"twentytwentyfive/cta-newsletter","name":"Newsletter Sign Up"},"align":"full","className":"is-style-section-3","style":{"spacing":{"padding":{"right":"var:preset|spacing|50","left":"var:preset|spacing|50","top":"var:preset|spacing|50","bottom":"var:preset|spacing|50"}},"dimensions":{"minHeight":""}},"layout":{"type":"constrained","contentSize":"800px"}} -->
```
carolinan: Nit: remove metadata.

carolinan (approved):
From my end, using the section style on the group works much better across all the combined style variations.

sudip-md:
### Test Report

**Description**:
There is still a Midnight style background without a contrast color for the text. Everything else looks good.
Suggested :  Midnight style (#4433A6)should have White font color.

Video - 

https://github.com/user-attachments/assets/1a063f02-9add-43cf-8fdf-8172bf13f99b

juanfra:
@sudip-md thank you. That's not how I see it in the branch where I worked. This is how I see it:

https://github.com/user-attachments/assets/5b2cdc7d-044e-4253-baae-043ac6940c1f

Can you please check that you're on the `update/newsletter-pattern` branch, have pulled the latest changes, and re-inserted the pattern once you pulled the changes?

patterns/cta-newsletter.php
juanfra: I believe we need to keep the categories so that the shuffle functionality can be more robust.

Have we concluded whether we're keeping the pattern name and name? I saw some discussions going on.

carolinan: But the category, slug, name everything is listed in the file header.
Who can give a definite answer?

juanfra: I could be getting it wrong, but it seems that the suffle functionality relies on the metadata. Please check [this code](https://github.com/WordPress/gutenberg/blob/03cfbff40ef2e5bf9ad369bb0a18642aa70833ef/packages/block-editor/src/components/block-toolbar/shuffle.js#L34).

carolinan: The metadata is inserted in the editors when you insert the pattern.
So it is not missing from the shuffle functionality. 

I went down a rabbit hole trying to find any official documentation, dev note or similar.

carolinan:
I think this can be merged but we need to define the questions clearly if we can find who to reach out to about the metadata, including the translation of the metadata.

patterns/cta-newsletter.php
juanfra: Sounds good, thanks for that. Let's try to define what we do with the metadata in the issue and work on it in bulk.

juanfra:
> I think this can be merged but we need to define the questions clearly if we can find who to reach out to about the metadata, including the translation of the metadata.

Sounds good. Shall we keep the conversation about metadata [here](https://github.com/WordPress/twentytwentyfive/issues/403)? I saw they pinged Daniel Richards and maybe he has some more insights about it.

juanfra:
I've removed the metadata and merging this one. In case we decide to add metadata to patterns we can revisit later in batch. Thank you 🙌

PR #530, opened 2024-10-09
Fix: Replace empty paragraph with spacer.

juanfra:
**Description**

Replace empty paragraph with spacer in cover blocks, so that the patterns have a better presentation in the editor.

| Before | After | 
| --- | --- |
| Screenshot 2024-10-09 at 11 07 09 | Screenshot 2024-10-09 at 11 08 30 |
| Screenshot 2024-10-09 at 11 05 27 | Screenshot 2024-10-09 at 11 04 09 |

**Testing Instructions**

1. Create a page.
2. Insert the following patterns: "Link in bio heading, paragraph, links and full-height image" and "Link in bio with tight margins"
3. Confirm there's no "Type / to choose a block" text in the images.


carolinan (approved):
Thank you, this is a good follow up to changes that have already been made in other cover blocks.

PR #563, opened 2024-10-11
Try: Update midnight & typography preset 7.

### LINKED ISSUE #549: Midnight: List of elements to adjust
While testing the **Midnight variation and its typography preset**, I noticed some elements could be improved:

Image
Image

Here are the changes that can improve the look of the variation & typeset:

- Post Author Name: reset styles
- Post Terms: reset styles
- Font size preset XXL: min `2rem`, max `2.8rem`
- Font size preset XL: min `1.8rem`, max `2.2rem`
- Font size preset L: min `1.2rem`, max `1.8rem`
- Font size preset M: min `0.9rem`, max `1.2rem`
- Font size preset S: `0.9rem`
- Text: set to font size preset M, letter spacing `-0,01em`, line height `1.5`
- Quote: reset font size (it should use the same preset as Text)
- Pullquote: remove specific line height
- Headings: line height `1.24`
- H6: remove specific font-weight (it should be the same as other headings, 300)
- Buttons: set to font size preset M, `font-weight: 400`, letter spacing `-0.01em`
- If it’s possible, can we remove Fira Sans Bold from the theme and whenever the Bold typography setting is selected and applied, it falls back to Fira Sans Medium? Mostly because Fira Sans Bold is appearing a bit harsh visually, probably because there’s no font smoothing, like there is on Figma.

How it's expected to look with the changes applied.

Image
Image
Image
Image

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/549

Updating the style variation and typography preset according to the request [here](https://github.com/WordPress/twentytwentyfive/issues/549)

**Screenshots**

- N/A

**Testing Instructions**

1. Go to the site editor
2. Pick the "Literata & Fira Sans" typographic preset and confirm the styles match the ones listed in the issue.

---

1. Go to the site editor
2. Pick the "Midnight" style variation and confirm the styles match the ones listed in the issue.

### DIFF
--- a/styles/08-midnight.json
+++ b/styles/08-midnight.json
@@ -95,7 +95,7 @@
 				{
 					"fluid": {
 						"max": "2.8rem",
-						"min": "2em"
+						"min": "2rem"
 					},
 					"name": "Extra Extra Large",
 					"size": "2rem",
@@ -108,7 +108,7 @@
 		"typography": {
 			"fontSize": "var:preset|font-size|medium",
 			"fontFamily": "var:preset|font-family|fira-sans",
-			"letterSpacing": "-0,01em",
+			"letterSpacing": "-0.01em",
 			"lineHeight": "1.5"
 		},
 		"blocks": {
@@ -203,7 +203,7 @@
 			"core/quote": {
 				"typography": {
 					"fontSize": "var:preset|font-size|medium",
-					"letterSpacing": "-0,01em",
+					"letterSpacing": "-0.01em",
 					"lineHeight": "1.5",
 					"fontWeight": "300"
 				}
--- a/styles/08-midnight.json
+++ b/styles/08-midnight.json
@@ -62,12 +62,12 @@
 				{
 					"fluid": false,
 					"name": "Small",
-					"size": "0.875rem",
+					"size": "0.9rem",
 					"slug": "small"
 				},
 				{
 					"fluid": {
-						"max": "1.1rem",
+						"max": "1.2rem",
 						"min": "0.9rem"
 					},
 					"name": "Medium",
@@ -76,29 +76,29 @@
 				},
 				{
 					"fluid": {
-						"max": "1.2rem",
-						"min": "1.09rem"
+						"max": "1.8rem",
+						"min": "1.2rem"
 					},
 					"name": "Large",
-					"size": "1.38rem",
+					"size": "1.2rem",
 					"slug": "large"
 				},
 				{
 					"fluid": {
-						"max": "2rem",
-						"min": "1.75rem"
+						"max": "2.2rem",
+						"min": "1.8rem"
 					},
 					"name": "Extra Large",
-					"size": "1.75rem",
+					"size": "1.8rem",
 					"slug": "x-large"
 				},
 				{
 					"fluid": {
-						"max": "3.625rem",
-						"min": "2.625rem"
+						"max": "2.8rem",
+						"min": "2em"
 					},
 					"name": "Extra Extra Large",
-					"size": "2.625rem",
+					"size": "2rem",
 					"slug": "xx-large"
 				}
 			]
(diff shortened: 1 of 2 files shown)


styles/08-midnight.json
```diff
@@ -76,39 +76,40 @@
...
 			]
 		}
 	},
 	"styles": {
 		"typography": {
+			"fontSize": "var:preset|font-size|medium",
 			"fontFamily": "var:preset|font-family|fira-sans",
-			"letterSpacing": "-0.24px",
-			"lineHeight": "1.4"
+			"letterSpacing": "-0,01em",
```
carolinan: The comma makes the value invalid.

styles/08-midnight.json
```diff
@@ -76,39 +76,40 @@
...
-					"size": "1.75rem",
+					"size": "1.8rem",
 					"slug": "x-large"
 				},
 				{
 					"fluid": {
-						"max": "3.625rem",
-						"min": "2.625rem"
+						"max": "2.8rem",
+						"min": "2em"
```
carolinan: Question: is this supposed to be rem or em?

styles/08-midnight.json
```diff
@@ -218,6 +202,9 @@
 			},
 			"core/quote": {
 				"typography": {
+					"fontSize": "var:preset|font-size|medium",
+					"letterSpacing": "-0,01em",
```
carolinan: Same here, change out the comma for the period.

styles/typography/typography-preset-7.json
```diff
@@ -112,16 +99,23 @@
...
 					"fontFamily": "var:preset|font-family|literata",
 					"fontWeight": "200",
-					"letterSpacing": "-0.02em"
+					"letterSpacing": "-0.02em",
+					"lineHeight": "1.24"
+				}
+			},
+			"h6": {
+				"typography": {
+					"fontWeight": "200"
```
carolinan, line 116: The H6 already inherits the 200 font weight set on elements > heading.

styles/typography/typography-preset-7.json
```diff
@@ -112,16 +99,23 @@
 			"button": {
 				"typography": {
 					"fontFamily": "var:preset|font-family|literata",
-					"fontWeight": "600",
-					"letterSpacing": "-0.96px",
+					"fontSize": "var:preset|font-size|medium",
```
carolinan: The font size is already set to medium in theme.json > elements > button, and it is the same as on the body, so this can be removed.

styles/typography/typography-preset-7.json
```diff
@@ -112,16 +99,23 @@
...
+					"fontWeight": "400",
+					"letterSpacing": "-0.01em",
```
carolinan: Same here, since the letter spacing would be the same as on the body, it does not need to be duplicated.

styles/08-midnight.json
carolinan: Actually, same here as with the buttons: Values that are identical to do those on the body does not need to be duplicated on the block.

juanfra:
Thank you Carolina!

> The H6 already inherits the 200 font weight set on elements > heading.

Given the specificity of [this](https://github.com/WordPress/twentytwentyfive/blob/trunk/theme.json#L1438), the heading styles are being overridden, and we need to set the H6.

carolinan:
> Given the specificity of [this](https://github.com/WordPress/twentytwentyfive/blob/trunk/theme.json#L1438), the heading styles are being overridden, and we need to set the H6.

I was able to confirm this, good catch.

beafialho:
Thank you, it's all looking good to me 👍

> If it’s possible, can we remove Fira Sans Bold from the theme and whenever the Bold typography setting is selected and applied, it falls back to Fira Sans Medium? Mostly because Fira Sans Bold is appearing a bit harsh visually, probably because there’s no font smoothing, like there is on Figma.

I assume this isn't possible. If that's the case, it's ok, it's not a blocker.

juanfra:
Thank you all! as we discussed in Slack we keep the bold. I'm merging this.

Issue #538, opened 2024-10-09
Afternoon: List of elements to adjust


beafialho:
While testing **Afternoon variation and its typography preset**, I noticed some size issues that make it look bad.

![Image](https://github.com/user-attachments/assets/150c6fbd-58e1-45f7-8402-8f9284be396b)

I tried to edit the files myself but the chance of messing them up was high, so I tried my best to enumerate the following changes I made in Global Styles in the editor, which I kindly ask be taken to a PR.

- Change font preset XXL: min `2rem`, max `2.6rem`
- Change font preset XL: min `1.4rem`, `1.8rem`
- Text: set to `font-weight: 300`, line height to `1,5`
- Headings: set to `font-weight: 400`, letterSpacing: -0,1px , line-height: 1.2
- H1: XXL font size
- H2: XL font size
- H3: L font size
- H4: M font size
- H5: M font size
- H6: S font size + (remove uppercase, letter spacing, remove specific font weight, it should be `400` same as others)
- Site Title: set to Large font size
- Post author, post author name, post date, post terms: remove font size preset added globally
- Titles (Post Titles): any specific settings should be removed, they should inherit the headings settings

Here's how it's supposed to look with these changes:

![Image](https://github.com/user-attachments/assets/e6fc1794-47c0-4856-b9a9-4b8d3a8fa187)
![Image](https://github.com/user-attachments/assets/f0c11572-38d2-45ec-a59b-175300a728c0)
![Image](https://github.com/user-attachments/assets/de8bd91e-e1d2-4ba9-ae10-066c3f2bf54a)
![Image](https://github.com/user-attachments/assets/f551f4bb-fb6f-4af3-92e1-87bd5935576d)
![Image](https://github.com/user-attachments/assets/57fc8208-4a20-47af-ae37-613e467135fb)

cc: @juanfra @carolinan @richtabor

carolinan:
Is it intentional the H4 and H5 should be identical? I am not sure that is ideal.

beafialho:
Yes, it is intentional. I assumed we'd prefer to use font size presets in the headings.

carolinan:
I wish I could get 100 likes on https://github.com/WordPress/gutenberg/issues/65108

troychaplin:
@carolinan @beafialho I just submitted PR #564 to fix some of the items in the original list

> Change font preset XXL: min 2rem, max 2.6rem
> Change font preset XL: min 1.4rem, 1.8rem
[...]

PR #564, opened 2024-10-11
Update/afternoon presets

troychaplin:
**Description**

Fixing some of the items noted in #538

### DIFF
--- a/styles/04-afternoon.json
+++ b/styles/04-afternoon.json
@@ -149,16 +149,6 @@
 					"fontSize": "var:preset|font-size|medium"
 				}
 			},
-			"core/post-author": {
-				"typography": {
-					"fontSize": "var:preset|font-size|medium"
-				}
-			},
-			"core/post-author-biography": {
-				"typography": {
-					"fontSize": "var:preset|font-size|medium"
-				}
-			},
 			"core/post-terms": {
 				"typography": {
 					"fontWeight": "400"
--- a/styles/04-afternoon.json
+++ b/styles/04-afternoon.json
@@ -97,7 +97,7 @@
 	"styles": {
 		"typography": {
 			"fontFamily": "var:preset|font-family|ysabeau-office",
-			"fontWeight": "400",
+			"fontWeight": "300",
 			"letterSpacing": "-0.22px",
 			"lineHeight": "1.6"
 		},
--- a/styles/04-afternoon.json
+++ b/styles/04-afternoon.json
@@ -99,7 +99,7 @@
 			"fontFamily": "var:preset|font-family|ysabeau-office",
 			"fontWeight": "300",
 			"letterSpacing": "-0.22px",
-			"lineHeight": "1.6"
+			"lineHeight": "1.5"
 		},
 		"blocks": {
 			"core/button": {
--- a/styles/04-afternoon.json
+++ b/styles/04-afternoon.json
@@ -169,12 +169,6 @@
 					"fontWeight": "400"
 				}
 			},
-			"core/post-title": {
-				"typography": {
-					"fontWeight": "500",
-					"letterSpacing": "-0.8px"
-				}
-			},
 			"core/pullquote": {
 				"typography": {
 					"fontFamily": "var:preset|font-family|platypi",
--- a/styles/04-afternoon.json
+++ b/styles/04-afternoon.json
@@ -97,7 +97,6 @@
 	"styles": {
 		"typography": {
 			"fontFamily": "var:preset|font-family|ysabeau-office",
-			"fontWeight": "300",
 			"letterSpacing": "-0.22px",
 			"lineHeight": "1.5"
 		},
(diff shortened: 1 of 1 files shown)


karmatosed:
@troychaplin it looks like a few things would be helpful here:

1. Can you please provide all the fixes if possible for this in the PR so we can get them all in? Unless I am mistaken some are missing.
2. Can you show a screenshot of the changes please?

troychaplin:
> @troychaplin it looks like a few things would be helpful here:
> 
>     1. Can you please provide all the fixes if possible for this in the PR so we can get them all in? Unless I am mistaken some are missing.
> 
>     2. Can you show a screenshot of the changes please?

More than happy to try, I made some comments on the [issue here](https://github.com/WordPress/twentytwentyfive/issues/538), would be happy to cycle back and try and finish them off with a bit of clarity on my notes. Might not get a chance to look until tomorrow, holiday weekend in Canada, away most of the day.

troychaplin:
@karmatosed here are some before and after images of each change I've made.

--
Change font preset XXL: min 2rem, max 2.6rem & Change font preset XL: min 1.4rem, 1.8rem

font-preset-xxl-xl

--
Text: set to font-weight: 300, line height to 1,5

text-weight-height

--
Headings: set to font-weight: 400, letterSpacing: -0,1px , line-height: 1.2

heading-weight-space-height

--
Site Title: set to Large font size

site-title-large

--
Titles (Post Titles): any specific settings should be removed, they should inherit the headings settings

post-titles-remove-settings

troychaplin:
@karmatosed In the json file for afternoon I changed the styles for `post-author-name` but now I wonder if `post-author` should also be changed? (maybe even `post-author-biography`?)

--
Another thing I noticed and want to make sure it's of no concern. When I remove the `post-author-name` in the afternoon json file it displays larger than I would expect in the style book, like the following:

removed-from-json

--
If I change it to small instead of removing it, the size displays correctly in the style book like this:

set-as-small

--
Regardless of the size display in the book, the front end is correct regardless of which change is made, I test both and they both look like the following:

frontend-output

troychaplin:
@karmatosed I went back to the list of heading sizes and I think it clicked. I tested that each heading was set to the size in the original list by temporary changing the `fontSizes` under `settings` / `typography` and testing in the editor and style book.

In the process I added settings for h5 and h6 as per the original notes in issue #538 

The original list of requested changes are:

```
H1: XXL font size
H2: XL font size
H3: L font size
H4: M font size
H5: M font size
H6: S font size + (remove uppercase, letter spacing, remove specific font weight, it should be 400 same as others)
```

Adding styles for h5 and h6 looks like the following:

heading-sizes

carolinan:
Correct, if the sizes match those that are in theme.json, they do not need to be added in these files, they will be inherited from theme.json.

beafialho:
Thank you @troychaplin!

All is looking as expected to me.

> In the json file for afternoon I changed the styles for post-author-name but now I wonder if post-author should also be changed?

That would be ok, I think. I don't think that block is used in the templates, but if it has specific settings it could potentially behave awkwardly if users add it to templates so maybe that's a wise call.

troychaplin:
> Thank you @troychaplin!
> 
> All is looking as expected to me.
> 
> > In the json file for afternoon I changed the styles for post-author-name but now I wonder if post-author should also be changed?
> 
> That would be ok, I think. I don't think that block is used in the templates, but if it has specific settings it could potentially behave awkwardly if users add it to templates so maybe that's a wise call.

Ok, will make this last change shortly!

troychaplin:
@karmatosed @beafialho @carolinan I have removed `post-author`. I also took a look at `post-author-biography` and also remove it and now all the `post-*` referenced in the style book are consistent

styles/04-afternoon.json
```diff
@@ -97,9 +97,9 @@
 	"styles": {
 		"typography": {
 			"fontFamily": "var:preset|font-family|ysabeau-office",
-			"fontWeight": "400",
+			"fontWeight": "300",
```
juanfra: This value is already coming from `theme.json`

```suggestion
```

styles/04-afternoon.json
```diff
@@ -132,27 +132,19 @@
...
-			"core/list": {
+			"core/heading": {
 				"typography": {
-					"lineHeight": "1.3"
-				}
-			},
-			"core/loginout": {
-				"typography": {
-					"fontSize": "var:preset|font-size|medium"
+					"fontWeight": "400",
```
juanfra: This value is already coming from `theme.json`

```suggestion
```

styles/04-afternoon.json
```diff
@@ -132,27 +132,19 @@
...
+					"letterSpacing": "-0.1px",
```
juanfra: This value is already coming from `theme.json`
```suggestion
```

juanfra (commented):
Thank you very much @troychaplin - Looking good to me.

I left a few adjustments I'm going to commit now and then we're good to merge.

juanfra (approved):

PR #568, opened 2024-10-14
Morning: Update the color contrast on the button in section style 4

### LINKED ISSUE #492: Test button color contrast

The button block and button element has CSS applied that changes the background color on hover. 
It mixes 85% of the background color with transparent: ```background-color: color-mix(in srgb, var(--wp--preset--color--contrast) 85%, transparent);```
But it leaves the text color unchanged.

The result is that the contrast ratio between the button background and the button text color changes depending on what the surrounding background color is.

The color change is not used as the only indicator that a button is being hovered.
But the contrast ratio between the button background color and the text color must be high enough for the text to still be readable.
For simplicity's sake let's assume that the text size is the default 18px and that the minimum contrast ratio is 4.5:1 (WCAG AA).

-------
The theme has 9 theme style variations including the default:

Image alt="Styles sidebar with theme style variations visible"

Every style variation has 6 section styles:
Image alt="The section style options are shown in the block settings of the group block"

In short, these options are used to quickly change the colors of a group block.
How this works is that the users inserts one of the patterns, or, they add a group block themselves,
selects one of the 6 styles, and places a button inside the group.

What needs to be tested is that the button colors are still accessible when these section style options are used, in each theme style variation.
What makes this so time consuming is that I found that automated contrast checking tools can not check the color contrast when color-mix is used.

Example:

1. Go to Appearance > Editor. Select any template to edit.
2. Open the Styles sidebar and reset all saved style changes.
3. Insert a group block.
4. Insert two buttons: one default and one outlined. Remember to add the links to the buttons.
5. Test the color contrast.
6. Select the group block again. In the block settings sidebar, open the Styles tab.
7. In the Styles panel, select style 2 and test the color contrast. Repeat this for all 6 styles.

Repeat from step 2, for each theme style variation in the Styles sidebar > Browse styles.

carolinan:
**Description**
This PR updates the text color on the button element in the Morning variation when the button is placed in section style 4, and the button is hovered.
It overrides the text color that is set in styles/sections/section-4.json.

The resulting contrast ratio between the button text and the button text color is 5.08:1
Closes https://github.com/WordPress/twentytwentyfive/issues/492

**Screenshots**
On hover:
image

**Testing Instructions**

Go to Appearance > Editor > Styles and enable "Morning".
Create a new post.
Insert a group block and apply section style 4.
In the group, insert a buttons block with a filled button.
Confirm that when you hover over the button, the text color is set to `contrast`.

### DIFF
--- a/styles/06-morning.json
+++ b/styles/06-morning.json
@@ -302,9 +302,10 @@
 						},
 						":hover": {
 							"color": {
-								"background": "color-mix(in srgb, var(--wp--preset--color--accent-1) 85%, transparent)"
+								"background": "color-mix(in srgb, var(--wp--preset--color--accent-1) 85%, transparent)",
+								"text": "var:preset|color|contrast"
+							}
 						}
-					}
 					},
 					"link": {
 						"color": {


juanfra (approved):
Thank you, works well!

https://github.com/user-attachments/assets/4bf2f905-8954-46b2-8f6a-4f5d4372fb1a

Issue #404, opened 2024-09-25
Visible gap between the footer and the Newsletter signup pattern


carolinan:
**Description**
The newsletter signup pattern is placed as the last section in 3 designs:
- News blog with featured posts grid
- Landing page for Book
- Landing page for Podcast

In the news blog template, and when the two landing pages uses either of the footers that has a solid color background, there is a visible gap between the footer and pattern.

**Step-by-step reproduction instructions**
Go to Appearance > Editor > Templates and select the Home template.
In the Design panel, select the News blog with featured posts grid, and save. 
Confirm that there is a gap between the pattern and footer.

Go back to Templates and select Page.
Replace the default footer with one of the footers that have a solid color background.
Save.
Create a new page. In the pattern selection modal, select "Landing page for Book", and save.
View the footer area of the page on the front.
Repeat for "Landing page for Podcast"

**Expected behavior**

In my _opinion_ the sections would look better without this gap.

**Screenshots**
The landing pages with the black footer:
![image](https://github.com/user-attachments/assets/56b7f84b-f9ba-40ae-a7f9-8338ba52f2bb)

News blog with featured posts grid:
![image](https://github.com/user-attachments/assets/b17efa4d-f1bc-4019-b78a-1740ffe54b64)

alaminfirdows:
**template: page-no-title**

It comes from `.wp-site-blocks`. Should we reset the margin like we do for the main section (post content)?

![image](https://github.com/user-attachments/assets/55edef88-e2db-4ee1-8630-2ce0f5a90244)

alaminfirdows:
**template: page**

It includes padding for post content, in addition to the margin from `.wp-site-blocks`.

![image](https://github.com/user-attachments/assets/8f64327d-7c4a-4432-a8bf-4c2d6ddc26ff)

carolinan:
The problem is that if the top margin is removed from the footers with colored backgrounds, it probably needs to be removed from all footers, and the spacing between the footers and content is going to be affected.
A decision needs to be made about which is preferable.

beafialho:
Correct, the footers should probably have no margin so that this doesn't happen with coloured sections 👍

carolinan:
This is affected by the following issue:
There is no way to remove the default top margin on the footer template part, without wrapping the part inside a group or using custom CSS:
https://github.com/WordPress/gutenberg/issues/47637

PR #572, opened 2024-10-14
Remove the top margin from the footer template part

carolinan:
**Description**
The purpose of this change is to remove the spacing between the footer template part and full width, color patterns placed above it,

Partial for https://github.com/WordPress/twentytwentyfive/issues/404

**Screenshots**
Before:
370705205-b17efa4d-f1bc-4019-b78a-1740ffe54b64

After
Partial screenshot just to show that the gap is gone:
image
(The size difference here is not related to the PR)

**Testing Instructions**
Create a new page. 
Select "Landing page for Book". 
Confirm that there is no gap between the footer and the last pattern.

Go to Appearance > Editor > Templates and select the Home template.
In the Design panel, select the News blog with featured posts grid, and save.
Confirm that there is no gap between the footer and the last pattern.

### DIFF
--- a/assets/css/editor-style.css
+++ b/assets/css/editor-style.css
@@ -6,3 +6,11 @@ a {
 	text-decoration-thickness: 1px !important;
 	text-underline-offset: .1em;
 }
+
+/*
+ * Remove the top margin from the footer template parts.
+ * https://github.com/WordPress/gutenberg/issues/47637
+ */
+footer.wp-block-template-part {
+	margin-block-start: 0;
+}
--- a/style.css
+++ b/style.css
@@ -44,3 +44,11 @@ h1, h2, h3, h4, h5, h6, blockquote, caption, figcaption, p {
 .more-link {
 	display: block;
 }
+
+/*
+ * Remove the top margin from the footer template parts.
+ * https://github.com/WordPress/gutenberg/issues/47637
+ */
+footer.wp-block-template-part {
+	margin-block-start: 0;
+}


juanfra (commented):
Thanks Carolina. 

I still see a gap for the first scenario.

https://github.com/user-attachments/assets/5612dc3a-3118-4073-9e56-f725afb4f336

For the second scenario, it works well with that particular home template, but with others it doesn't. Not sure if we want to address that in a separate PR


https://github.com/user-attachments/assets/a89c41d7-9b4d-4488-b084-da6d6f12d2b1

carolinan:
Can we remove the newsletter-signup pattern from this landing page instead? Is it not strange to have two signup buttons this close?

Edit:
Sorry for being unclear, I meant removing it from the "News blog with featured posts grid" pattern, not the landing page patterns. The landing page patterns only shows two "Subscribe" buttons if the user also updates the footer.

juanfra:
> Can we remove the newsletter-signup pattern from this landing page instead? Is it not strange to have two signup buttons this close?

I wouldn't be opposed because it would solve this issue. I'll defer to @beafialho as she may want to keep the design as is for that section.

beafialho:
> Can we remove the newsletter-signup pattern from this landing page instead? Is it not strange to have two signup buttons this close?

The [original design](https://www.figma.com/design/dzGCSntVch4EQdVERTqyVK/Twenty-Twenty-Five?node-id=3137-3091&t=66DtOenIeIVNJ8O8-4) of this template shows a different footer, so I suggest using the intended footer.

News blog with featured posts grid

carolinan:
I will create a new part for the footer-columns pattern. Can you check if there are any other templates that have the wrong footer?

beafialho:
> I will create a new part for the footer-columns pattern. Can you check if there are any other templates that have the wrong footer?

Yes, from what I'm seeing the only other template that should also use this footer is "News blog with sidebar".

carolinan:
@beafialho I just want to make sure: The News blog with sidebar should be updated to use the footer with columns?

beafialho:
yes @carolinan, "News blog with sidebar" according to the design, should use the "Footer with columns". For clarification this is the intended appearance:

There's one thing I'd like clarification on, though: can we/do we want to set the headers and footers used in the Figma designs to each template? If that's the case, I can open up an issue to address the ones in News blog templates, because some of them are not using the headers and footers specified in the design.

News blog with sidebar

carolinan:
We can, I think we have time to do this before Monday evening (RC1)

beafialho:
> We can, I think we have time to do this before Monday evening (RC1)

Thank you for clarifying. None of this is critical, it is fine-tuning. I have opened an issue [here](https://github.com/WordPress/twentytwentyfive/issues/591).

Issue #571, opened 2024-10-14
Replacing the copyright block binding


carolinan:
**Is your feature request related to a problem? Please describe.**
Cross posting this ticket here so that it is not missed
https://core.trac.wordpress.org/ticket/62143

I believe the actionable points for now are:

- Replace the copyright footer text with the theme name

carolinan:
What about keeping the copyright date binding in the theme, but letting the user decide if they want to use it?
And change the date to the date of the first published post or page, not the current year.

2ndkauboy:
I read the conversation about this issue. I believe the issue with "naming the copyright holder" would still persist, right? Then not using a "copyright footer text" but only the theme name might be best.

carolinan:
@beafialho Are there any places in the design that would benefit from only displaying one category or tag?

With the post terms block, there is no way to limit how many categories or tags that are showing.
With the block bindings, we could limit it to one. The problem is that the user would not be able to select _which_ single category or tag that shows.

In the editor the placeholder text could be something like "first category".

beafialho:
> [@beafialho](https://github.com/beafialho) Are there any places in the design that would benefit from only displaying one category or tag?
> 
> With the post terms block, there is no way to limit how many categories or tags that are showing. With the block bindings, we could limit it to one. The problem is that the user would not be able to select _which_ single category or tag that shows.
> 
> In the editor the placeholder text could be something like "first category".

Interesting question @carolinan. Perhaps the News blog templates would be good places to try (only the Blog templates, not the Single post templates though). Given that these designs contemplate one category for each post as a way to reduce the amount of information on the homepage.

Maybe:
- News blog home
- News blog with featured posts grid
- News blog with sidebar

carolinan:
I have not tried it inside a query loop, will check if my test code works.

carolinan:
@beafialho @juanfra I tried using the block binding to display the category on these templates, but the problem is that it does not inherit the style from the terms block, since it uses a paragraph.
So when I switch style variations, the letter spacing and font weight do not match the current styles.

I also found that in the editor, where the placeholder shows, the styling is also different because there is no link to the category archive. The link is only used on the front.

juanfra:
Thanks Carolina, I could see that. As it's related to aesthetics, I'll defer to @beafialho and her opinion on how that would coexist in her design.

PR #574, opened 2024-10-15
Remove the block binding

carolinan:
**Description**
This PR removes the copyright block binding because the format of the copyright notice is incorrect.
The copyright notice in the footers is replaced with the theme name.

Partial for https://github.com/WordPress/twentytwentyfive/issues/571

**Testing Instructions**

Confirm that the theme name is showing at the bottom left of
- The default footer
- Footer with columns
- Footer with newsletter signup

And in the Portfolio homepage template.


juanfra (commented):
Thanks for the changes, works well! The only thing I'm doubtful is if we have to pass `Twenty Twenty-Five` to translation functions. I believe, just like `WordPress`, these stay the same in every language and that's the intention.

https://github.com/user-attachments/assets/b48799dc-ec1f-4270-95ad-104b0d61a76a

carolinan:
The way it was explained to me many many years ago was that it needs to be translation ready because the script is different. WordPress in for example Arabic is `ووردبريس`  but an URL like wordpress.org is wordpress.org

2ndkauboy:
Wouldn't it make more sense to use the "Site Title" instead of the theme name? Only in some footers we would have the site title twice then.

I don't really think the theme name - with no link - does make sense for most users. The site name kind of makes sense.

Or just don't have the whole "left column" at all, if it would duplicate the site title? :thinking:

carolinan:
I agree with using the site title, or even using the site tagline, but I know that @beafialho asked for the theme name.

carolinan:
In my opinion if the left side was left empty it may look unbalanced

2ndkauboy:
@carolinan this is what I've meant: https://github.com/WordPress/twentytwentyfive/pull/578

juanfra:
> The way it was explained to me many many years ago was that it needs to be translation ready because the script is different. WordPress in for example Arabic is `ووردبريس` but an URL like wordpress.org is wordpress.org

Thanks for clarifying, Carolina!

Regarding having the site title, as mentioned, it was discussed and the problem is redundancy. It'd be displayed two times in a small piece of "real estate". 

> In my opinion if the left side was left empty it may look unbalanced

I fully agree with this.

If we want to discuss the modification of the footer, I'd suggest we open a new issue and discuss it there. @beafialho carefully crafted the design of the different footers (and the theme!), so I'd prefer if we ran any of these things with her first.

beafialho:
The footer was designed to have the site title (above) and the theme name showing discretely at the bottom (like many other themes do). I suggest we keep it like that.

Captura de ecrã 2024-10-16, às 10 13 58

beafialho:
We could however, for clarity, add `Twenty Twenty-Five Theme` instead of only `Twenty Twenty-Five`. I realise it may be somewhat abstract for users who aren't familiar with WordPress' default themes.

juanfra:
I'd personally prefer just `Twenty Twenty-Five`, but I could be biased as I believe `Twenty Twenty-Five Theme` wouldn't translate well to Spanish.

juanfra:
Thanks again Carolina, let's move forward with the theme name translated for now. And in case we need to make an adjustment we can revisit it later. Thank you all!

PR #575, opened 2024-10-15
Update the viewport width of the vertical header

carolinan:
**Description**
In the header and pattern previews in the Site Editor, the vertical header is very small.
The CSS transform that is used on the pattern preview also animates the pattern which makes it look like it moves from the top to bottom.

This pull request adds a viewport width to the vertical header pattern, which makes the preview zoom in on the design closer.
It also reduces the impact of the animation.

**Screenshots**
In these videos I am refreshing the Patterns data view screen in the Site Editor.

Before:

https://github.com/user-attachments/assets/bbd1ad3e-ee9a-4a8c-a170-e25e3bd06f8c

After:

https://github.com/user-attachments/assets/8e8105dd-c0a4-48b6-9ba5-133772d9e9d0

**Testing Instructions**
Go to Appearance > Editor > Patterns > Headers
Confirm that the vertical header is more visible than before.

### DIFF
--- a/patterns/vertical-header.php
+++ b/patterns/vertical-header.php
@@ -5,6 +5,7 @@
  * Categories: header
  * Block Types: core/template-part/vertical-header
  * Description: Vertical Header with site title and navigation
+ * Viewport width: 300
  *
  * @package WordPress
  * @subpackage Twenty_Twenty_Five


carolinan:
Note: the vertical header **template part** still shows the small pattern preview.

juanfra (commented):
I've tried following the steps described and I couldn't get it working.

https://github.com/user-attachments/assets/4b87b4ab-bdf9-4ef1-a899-57af3992141e

Not sure if I missed any step?

carolinan:
Select the header pattern, not the parts. The parts are not using the viewport widths, those are the HTML files.

carolinan:
https://github.com/WordPress/gutenberg/issues/66111

juanfra (approved):
Thanks for the clarification, I see the changes now. Works well.

https://github.com/user-attachments/assets/fe95c035-9c2a-441e-8ee3-5d41a4491fa7

PR #576, opened 2024-10-15
Try different block bindings

carolinan:
**Description**
This is for testing only.

I have registered two possible block bindings.
I have used patterns to display the binding to not limit where we can place them when testing.

The first displays the post format name.
Use case: To use in the single post view next to author/categories/etc, or in the post format patterns.
This binding only shows the post format name: "Link".  It could also be updated to use a prefix: "Format: Link".
If there is no format assigned, nothing shows.

The second displays the first category on the post, alphabetically.
Use case: Limiting the number of categories to show
This binding only shows the category name: "Apples".  It could also be updated to use a prefix: "Category: Apples".
The text is not linked, but could be.

> Perhaps the News blog templates would be good places to try (only the Blog templates, not the Single post templates though). Given that these designs contemplate one category for each post as a way to reduce the amount of information on the homepage.
>Maybe:
> News blog home
> News blog with featured posts grid
> News blog with sidebar

To test, create one or more posts with categories and or formats.
Insert the pattern where you want to test it, in a template, or content.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -157,7 +157,7 @@ if ( ! function_exists( 'twentytwentyfive_format_binding' ) ) :
 	function twentytwentyfive_format_binding() {
 		$post_format_slug = get_post_format();
 
-		if ( $post_format_slug && $post_format_slug !== 'standard' ) {
+		if ( $post_format_slug && 'standard' !== $post_format_slug ) {
 			return get_post_format_string( $post_format_slug );
 		}
 	}
--- a/functions.php
+++ b/functions.php
@@ -121,7 +121,7 @@ add_action( 'init', 'twentytwentyfive_pattern_categories' );
 // Registers block binding sources.
 if ( ! function_exists( 'twentytwentyfive_register_block_bindings' ) ) :
 	/**
-	 * Registers the copyright block binding source.
+	 * Registers the post format block binding source.
 	 *
 	 * @since Twenty Twenty-Five 1.0
 	 *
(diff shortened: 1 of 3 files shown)


juanfra:
Looks good. I see value in the post format one, until the functionality is fully fledged in core.

ironprogrammer:
Cool beans! I tried these out on both the blog home and single post templates, as well as inline for a post (I wasted no effort in design, obvs 😆). I think linking the category would be helpful for discovery of similar posts.

The bindings aren't "first class citizens" for using the inline inserter (`/foobar`), but come up with the regular "+" block inserter and pattern search.

- ✅ `twentytwentyfive/binding-format`: Displays post format when != `standard`.
- ✅ `twentytwentyfive/binding-category`: Displays first category alphabetically.

| Blog Home | Single Post |
| :----------: | :----------: |
|  |  |

carolinan:
Thank you for testing

dugyen:
**Try different block bindings**
**Description**
We created two different ways to link content to specific elements on a webpage. These links, called bindings, can be applied to any element, giving us flexibility in testing how they work in different situations.
**Environment**
- WordPress: 6.7-beta3
- PHP: 8.3.10
- Server: nginx/1.21.4
- Database: mysqli (Server: 5.7.44-log / Client: mysqlnd 8.3.10)
- Browser: Chrome 129.0.0.0
- OS: macOS
- Theme: Twenty Twenty-Five 1.0
- MU Plugins: None activated
- Plugins:
  * Gutenberg 19.4.0
  * Test Reports 1.1.0
  * WordPress Beta Tester 3.6.0

**Steps to Reproduce**
The first displays the post format name.
Use case: To use in the single post view next to author/categories/etc, or in the post format patterns.
This binding only shows the post format name: "Link". It could also be updated to use a prefix: "Format: Link".
If there is no format assigned, nothing shows.

The second displays the first category on the post, alphabetically.
Use case: Limiting the number of categories to show
This binding only shows the category name: "Apples". It could also be updated to use a prefix: "Category: Apples".
The text is not linked, but could be.

**Patch** 
1.  https://github.com/WordPress/twentytwentyfive/pull/576/files ✅ 

**Actual Results**
1.  Given patch is working ✅ 

**Supplemental Artifacts**
blog post

PR #578, opened 2024-10-15
Use site title instead of theme name in footer

2ndkauboy:
For those footers having the site title already, the additional block is removed.

https://github.com/user-attachments/assets/8c14830f-ca40-4bf8-97f7-897c125044f0


2ndkauboy:
Changes #574

beafialho:
Thanks for your suggestion. As mentioned [here](https://github.com/WordPress/twentytwentyfive/pull/574#issuecomment-2413717019) this significantly changes the design of the footers, adds redundancy and looks unbalanced.

2ndkauboy:
@beafialho I specifically tried not to have any redundancy by using the site title only in footer, where it is not already somewhere else (see video).

PR #581, opened 2024-10-16
Try: Fix display of the annotation.

### LINKED ISSUE #555: Potential problems with the flex CSS on the annotation text style
When the annotation text style is added to a paragraph or heading that is placed directly inside the post content, not nested in other blocks,

1. It is floating to the left of the content instead of being contained within the content width.
2. The text alignment options in the block toolbars stops working.
3.  Left and right margin options stops working

Without the flex CSS, the border around the block extends past the length of the text content itself, but all the options do work.

I would like to ask that it is carefully considered which of these scenarios is the best and less confusing for the user.

- Are there other CSS solutions?
- Can the theme add a block variation that disables these settings that do not work?

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/555

Changing the display so that it shows in context in the editor, and using a width to adjust to its content.

The alignment of the annotation will depend of its parent element (stack, for example).

**Screenshots**

https://github.com/user-attachments/assets/73964f09-0504-4a5a-b997-2a4c74c4264d

**Testing Instructions**

1. Create a page.
2. Add a paragraph of "annotation" style.
3. Confirm that the paragraph is located in context and not on the left as it was before.

### DIFF
--- a/styles/blocks/03-annotation.json
+++ b/styles/blocks/03-annotation.json
@@ -5,7 +5,7 @@
 	"slug": "text-annotation",
 	"blockTypes": ["core/heading", "core/paragraph"],
 	"styles": {
-		"css": "display: inline-flex",
+		"css": "width: fit-content; margin-inline: max(0px, ((100% - var(--wp--style--global--content-size)) / 2))",
 		"typography": {
 			"fontSize": "var:preset|font-size|small",
 			"lineHeight": "1.5",
--- a/styles/blocks/03-annotation.json
+++ b/styles/blocks/03-annotation.json
@@ -5,7 +5,7 @@
 	"slug": "text-annotation",
 	"blockTypes": ["core/heading", "core/paragraph"],
 	"styles": {
-		"css": "width: fit-content; margin-inline: max(0px, ((100% - var(--wp--style--global--content-size)) / 2))",
+		"css": "width: fit-content",
 		"typography": {
 			"fontSize": "var:preset|font-size|small",
 			"lineHeight": "1.5",


jasmussen:
Any reason to not include the !important? It makes the margin-inline still apply, which should not change, but in fact add to, the existing margins.

I'm happy to land this as a more solid alternative to the inline-flex that was applied before. I'd love a quick gut-check by @beafialho if she has time.

beafialho:
Thanks for the ping! Testing the behaviour, it seems to work nicely. I'm able to align it wherever I want to (left, right, inside columns, etc), even though the default alignment is centered.

juanfra:
> Any reason to not include the !important? It makes the margin-inline still apply, which should not change, but in fact add to, the existing margins.

@jasmussen the reason is that adding the important was pretty much affecting some layouts we already have in place, as I shared [here](https://github.com/WordPress/twentytwentyfive/issues/555#issuecomment-2414398748)


https://github.com/user-attachments/assets/22f05d19-98e0-4383-ac40-d492b2dec826

Recording the screencast, I realized that if we're ok with inheriting the margin rules, then we could remove the `margin-inline` and it will still work as I shared initially.

jasmussen:
Oh, gotcha. Then this PR is probably best to land as is! What do you think?

jasmussen (approved):

juanfra:
> Oh, gotcha. Then this PR is probably best to land as is! What do you think?

I believe we could also remove the `margin-inline: max(0px, ((100% - var(--wp--style--global--content-size)) / 2))` as it's not having an effect. Not sure if I'm missing some scenario I didn't think?

jasmussen:
Isn't that one what makes [Bea's alignments here work](https://github.com/WordPress/twentytwentyfive/pull/581#issuecomment-2416270944)? If not, and if it works without it, yes we can remove.

juanfra:
I believe the alignment will be coming from parent elements as things are, and the margin wouldn't have an effect. I pushed a commit removing the `margin-inline` for Bea to try it out. If it's ok we move forward without it, if not I revert.

@beafialho can you please try it again with the most recent changes?

Thank you both 🙌

mhkuu:
Hey there! 👋 Another test case is that with the current `display: inline-flex`, if you were to add any styling, that would look quite weird if your annotation spans multiple lines, as in the screenshot below:

Screenshot 2024-10-17 at 09 05 44

With `width: fit-content` as proposed in this solution, the layout is restored. Maybe an annotation spanning multiple lines including styling is not an intended use case for annotations but it's nice that this PR would solve the display of that as well.

juanfra:
Thanks for sharing Martijn. That's another good reason to get this one merged.

jasmussen:
Indeed! I'd suggest merging, so we can get some experience with it. Thanks for the work.

juanfra:
Perfect, merging then. Thank you all!

carolinan:
The text alignment still is not working.
I want to ask again if it would be better to create a block variation..?

carolinan:
The text alignment works if I enable width: "fit", but I am not sure users will understand how these work together.

jasmussen:
The text alignment is visible if you a full paragraph. The fact that it's centered is a bit of a tradeoff, that I think is acceptable in keeping the style as it is. It allows patterns to work, it's ergonomic, and arguably centering the pill-shape can also be an aspect of how it works. I would recommend keeping as-is.

carolinan:
> The text alignment is visible if you a full paragraph.

I don't understand. I think a word is missing?

jasmussen:
Ack, my apologies, chaotic morning. I meant to say: "The text alignment is visible if you write a full paragraph", by which I mean this:

Screenshot 2024-10-18 at 09 42 32

I know that's not the intent of the text style. But it matters to me that on a technical level, the text alignment isn't actually broken, it's just not visible when the item is centered. It's similar to any paragraph you put inside a Row block:

Screenshot 2024-10-18 at 09 43 35

Because of this, it still feels reasonable to offer this text style. You can try it out when you're authoring your docs, see if it fits. And it provides better ergonomics once you edit some of the TT5 patterns, which may be the more prevalent use case of the pill shape.

carolinan:
OK, thank you.

PR #590, opened 2024-10-18
Test category block binding in news blog template

carolinan:
**Testing Instructions**

Go to Appearance > Editor > Templates
Select Blog home
In the Design panel, select  News blog home. Save

Go to the front of the site.
Look at the first post on the top left.
Confirm that it has two categories showing.
The top one, is the terms block. 
The second one is the new block binding.

Test the style of this block binding in different style variations.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -117,3 +117,44 @@ if ( ! function_exists( 'twentytwentyfive_pattern_categories' ) ) :
 	}
 endif;
 add_action( 'init', 'twentytwentyfive_pattern_categories' );
+
+// Registers block binding sources.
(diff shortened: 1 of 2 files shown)


carolinan:
The top one, "Uncategorized" is the terms block with its current styling.
The second one "First category" is the new block binding.

https://github.com/user-attachments/assets/99159f6d-ef43-40dc-aed9-3723855abde7

carolinan:
I think we are leaning towards not using this block binding? Which leaves us with either no binding, using the "post format name" or coming up with something else within the next few hours.

carolinan:
After discussing this with @juanfra  and during a ticket triage, I am closing this as "wont fix"

Issue #401, opened 2024-09-25
Reduce zip file size


t-hamano:
When I removed the files necessary for development from this repository and converted it to a zip file, I found that the file size in my environment was 14.711 MB.

The theme directory does not mention any [requirements](https://make.wordpress.org/themes/handbook/review/required/) regarding file size, but the maximum size of a zip file that can actually be uploaded to the theme directory is 10 MB:

https://wordpress.org/themes/upload/

![upload](https://github.com/user-attachments/assets/f8c4d777-cf13-4443-8f8c-7d86053e69a2)

Although there is a file size limit for the theme directory, it seems strange that the default theme exceeds that size.

One approach to reducing the size might be to convert to the AVIF image format proposed in #383.

The `/assets/fonts` directory (about 8 MB) and `/assets/images` directory (about 6 MB) take up most of the overall size.

Incidentally, the file sizes when downloading [the wordpress-develop theme directory](https://github.com/WordPress/wordpress-develop/tree/trunk/src/wp-content/themes) and converting each theme to zip are as follows:

| Theme Name          | Zip File Size |
| ------------------- | ------------- |
| Twenty Twenty-Four  | 2.896MB       |
| Twenty Twenty-Three | 2.160MB       |
| Twenty Twenty-Two   | 5.818MB       |
| Twenty Twenty-One   | 2.783MB       |
| Twenty Twenty       | 1.005MB       |
| Twenty Nineteen     | 0.620MB       |
| Twenty Seventeen    | 2.115MB       |
| Twenty Sixteen      | 1.807MB       |
| Twenty Fifteen      | 4.230MB       |
| Twenty Fourteen     | 2.173MB       |
| Twenty Thirteen     | 1.764MB       |
| Twenty Twelve       | 1.192MB       |
| Twenty Eleven       | 1.238MB       |
| Twenty Ten          | 1.097MB       |

carolinan:
I am not opposed to the using AVIF, core also has a script for reducing the image sizes that will be ran before commit.
I believe the font families are set though as part of the design and style variations.

t-hamano:
Now that this theme has shipped into core, I zipped [this directory](https://github.com/WordPress/wordpress-develop/tree/ff3bb8d6d08f4e3fa75c56d53a4843662b0014d7/src/wp-content/themes/twentytwentyfive) and measured its size, finding it to be **11.4MB**.

carolinan:
[...]

PR #603, opened 2024-10-21
Try variable fonts as a means to reduce package size

carolinan:
**Description**

Partial for https://github.com/WordPress/twentytwentyfive/issues/401.
By replacing the separate font files with a single variable font, the package size can be decreased somewhat, and the size of theme.json is decreased as well.

**Testing Instructions**

Test the typography preset that uses the updated font families:
Manrope
Fira code
Platypi
Vollkorn
Ysabeau Office

Check that any visual changes are acceptable, since there may be some minor differences in the two formats.
Conform that the correct font style settings are available in the Typography panel > Appearance


juanfra (commented):
Thank you Carolina, from what I tested it looks good.

beafialho:
Thank you for making this change! If my testing is showing accurately, these are looking similar to the previous ones to me. Just to confirm, do they look like the screenshots below to you as well? If so, this is ok to merge.

Captura de ecrã 2024-10-23, às 09 29 44
Captura de ecrã 2024-10-23, às 09 32 15
Captura de ecrã 2024-10-23, às 09 31 10

juanfra:
Yes, it is looking like the screenshots to me. I'm merging this one then.

juanfra:
Thank you all!

Issue #606, opened 2024-10-21
"Poster-like" section looking broken in the front-end


juanfra:
**Description**

When adding the poster-like section pattern in a page, and then viewing the page in the front-end (in large viewports > 1900px width), the text is breaking in several lines and the design looks broken.

**Step-by-step reproduction instructions**
1. Add a page.
2. Add the "Poster-like section" pattern.
3. View the page in a viewport that's bigger than 1900px width.

**Expected behavior**

**Screenshots**

![Image](https://github.com/user-attachments/assets/2f0b3993-9d09-460b-ad1e-ef6108327ad5)

troychaplin:
Column widths changing to 80 / 20 fixes this issue, I put in a PR for this in case it's an acceptable solution from a design point of view. Visually, nothing else changes.

juanfra:
Fixed via https://github.com/WordPress/twentytwentyfive/pull/611

PR #611, opened 2024-10-21
update: column widths on poster-like patter

troychaplin:
**Description**

This PR addresses #606 by changing the column widths to 80 and 20 in both instances of the block to allow for the larger font size in the example to properly fit at larger widths

**Screenshots**

example

**Testing Instructions**

1. Create new page
2. Add "Poster-like" pattern
3. Save and view front end at more than 1900px wide (I tested at 2056)

### DIFF
--- a/patterns/banner-poster.php
+++ b/patterns/banner-poster.php
@@ -18,16 +18,16 @@
 	<div class="wp-block-group alignwide" style="min-height:100vh">
 		<!-- wp:columns {"align":"wide","style":{"spacing":{"blockGap":{"top":"var:preset|spacing|50"}}}} -->
 		<div class="wp-block-columns alignwide">
-			<!-- wp:column {"width":"66.66%"} -->
-			<div class="wp-block-column" style="flex-basis:66.66%">
+			<!-- wp:column {"width":"80%"} -->
+			<div class="wp-block-column" style="flex-basis:80%">
 				<!-- wp:heading {"textAlign":"left","align":"wide","style":{"typography":{"fontSize":"12vw","lineHeight":"0.9","fontStyle":"normal","fontWeight":"300"}}} -->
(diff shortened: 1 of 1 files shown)


juanfra (commented):
Thanks Troy, it is working well to me. 

Adjusting the column width is one direction that could solve it, the other direction could be adjusting the font size. I'll defer to @beafialho to pick which way to go as both changes will affect the way things look in one way or the other.

https://github.com/user-attachments/assets/380be93d-cc07-4ee3-a3c7-0d96c68e95a5

beafialho:
Thank you, these changes are looking good.

One comment, where I'm seeing "Mon, Jan 1" on the top right corner, should be written "Aug 08—10 2025 Fuego Bar, Mexico City", like the designs:

Captura de ecrã 2024-10-22, às 15 22 39

juanfra:
Thanks, I added the changes for the text [here](https://github.com/WordPress/twentytwentyfive/pull/617) as we're almost at RC1 and we have the string freeze (hoping that we can make it).

troychaplin:
> Thanks, I added the changes for the text [here](https://github.com/WordPress/twentytwentyfive/pull/617) as we're almost at RC1 and we have the string freeze (hoping that we can make it).

Too late to get this in before the freeze now that you've taken care of the date issue?

juanfra:
@troychaplin for RC1 yes, which is being [released now](https://wordpress.slack.com/archives/C02RQBWTW/p1729612908266079). We needed to merge it ASAP, as RC1 comes with hard string freeze. This means that strings should be ready for translators at this point. We can merge these changes now and include them for RC2 next week.

You can check the entire release schedule [here](https://make.wordpress.org/core/6-7/).

juanfra (approved):
Code looks good, and Bea confirmed design looks well. Thanks Troy!

PR #615, opened 2024-10-22
Try: Update underline thickness

### LINKED ISSUE #612: Link underline thickness inconsistency across post terms and author
If you notice, post author name has a thicker underline than categories. This is happening in all style variations.

Image
Image
Image
Image

carolinan:
**Description**
An attempt to solve https://github.com/WordPress/twentytwentyfive/issues/612
I can clearly see the difference in the underline thickness in the screenshots in the issue, but I can't actually see it in the browser. I tried Chrome and Firefox on Windows. So I will need help determining if this change solves the problem.

**Testing Instructions**
View the default single post template in the Site Editor.
**It needs to be the Site Editor**, because in this editor, the category in the post terms is not a link.
Below the featured image, observer the thickness of the underlines on the post author name and the category.
The thickness should be the same, 1px.

### DIFF
--- a/assets/css/editor-style.css
+++ b/assets/css/editor-style.css
@@ -2,7 +2,8 @@
  * Link styles
  * https://github.com/WordPress/gutenberg/issues/42319
  */
-a {
+a,
+.wp-block[style*="text-decoration: underline"] {
 	text-decoration-thickness: 1px !important;
 	text-underline-offset: .1em;
 }


richtabor:
I don't think we should be applying such specific styling. Instead we should avoid styling regular text as links. 

If a user does apply an underline on a linked text, then having it larger (currently in trunk) is fine, as it helps you identify the link.

carolinan:
Hopefully the bug with the missing <a> element on the category in the Site Editor can be solved soon, I saw a similar PR in the Gutenberg repository earlier today.

PR #623, opened 2024-10-22
Add translator note for multilingual strings.

peterwilsoncc:
**Description**

* adds a translator note for the strings `“Stories, historias, iсторії, iστορίες”` to clarify that the strings are intended to contain the word "Stories" in four languages.
* Modifies the context for each of the sample/placeholder headings to be consistent. In one location it differed.

**Screenshots**

N/A

**Testing Instructions**

1. Ensure the addition of the line breaks within the headings has not impacted the display negatively.

### DIFF
--- a/patterns/banner-poster.php
+++ b/patterns/banner-poster.php
@@ -24,7 +24,7 @@
 				<h2 class="wp-block-heading alignwide has-text-align-left" style="font-size:12vw;font-style:normal;font-weight:300;line-height:0.9">
 					<?php
 						/* translators: This string contains the word "Stories" in four different languages with the first item in the locale's language. */
-						echo esc_html_x( '“Stories, historias, iсторії, iστορίες”', 'Sample heading in four languages.', 'twentytwentyfive' );
+						echo esc_html_x( '“Stories, historias, iсторії, iστορίες”', 'Placeholder heading in four languages.', 'twentytwentyfive' );
 					?>
 				</h2>
 				<!-- /wp:heading -->
--- a/patterns/banner-poster.php
+++ b/patterns/banner-poster.php
@@ -23,8 +23,10 @@
 				<!-- wp:heading {"textAlign":"left","align":"wide","style":{"typography":{"fontSize":"12vw","lineHeight":"0.9","fontStyle":"normal","fontWeight":"300"}}} -->
 				<h2 class="wp-block-heading alignwide has-text-align-left" style="font-size:12vw;font-style:normal;font-weight:300;line-height:0.9">
 					<?php
+					echo wp_kses_post(
 						/* translators: This string contains the word "Stories" in four different languages with the first item in the locale's language. */
-						echo esc_html_x( '“Stories, historias, iсторії, iστορίες”', 'Placeholder heading in four languages.', 'twentytwentyfive' );
+						_x( '“Stories, <span lang="es">historias</span>, <span lang="uk">iсторії</span>, <span lang="el">iστορίες</span>”', 'Placeholder heading in four languages.', 'twentytwentyfive' )
+					);
 					?>
 				</h2>
 				<!-- /wp:heading -->
--- a/patterns/banner-poster.php
+++ b/patterns/banner-poster.php
@@ -21,7 +21,12 @@
 			<!-- wp:column {"width":"80%"} -->
 			<div class="wp-block-column" style="flex-basis:80%">
 				<!-- wp:heading {"textAlign":"left","align":"wide","style":{"typography":{"fontSize":"12vw","lineHeight":"0.9","fontStyle":"normal","fontWeight":"300"}}} -->
-				<h2 class="wp-block-heading alignwide has-text-align-left" style="font-size:12vw;font-style:normal;font-weight:300;line-height:0.9"><?php echo esc_html_x( '“Stories, historias, iсторії, iστορίες”', 'Sample heading in four languages.', 'twentytwentyfive' ); ?></h2>
+				<h2 class="wp-block-heading alignwide has-text-align-left" style="font-size:12vw;font-style:normal;font-weight:300;line-height:0.9">
+					<?php
+						/* translators: This string contains the word "Stories" in four different languages with the first item in the locale's language. */
+						echo esc_html_x( '“Stories, historias, iсторії, iστορίες”', 'Sample heading in four languages.', 'twentytwentyfive' );
+					?>
+				</h2>
 				<!-- /wp:heading -->
 			</div>
 			<!-- /wp:column -->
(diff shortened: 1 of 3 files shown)


juanfra (commented):
Thank you Peter! The addition of the line breaks in comments don't seem to be impacting the display.

https://github.com/user-attachments/assets/eb08a4fe-c89b-48c2-b622-32d68e56c0f9

I have a question regarding the translators comment. What would the expectation be for translators with the following part of the comment: "with the first item in the locale's language."

If I'd be translating to Spanish and I'd read that comment I'd think that the translated string should be" `“Historias, stories, iсторії, iστορίες“`. In French `“Histoires, historias, iсторії, iστορίες“`.

Should we give the direction of having the first word in the locale's language or should we leave it open so that the translators take care of it? I am asking this as I am not sure how this would translate to other languages in some locales and if it'd be affecting the display of some patterns.

peterwilsoncc:
> Should we give the direction of having the first word in the locale's language or should we leave it open so that the translators take care of it? I am asking this as I am not sure how this would translate to other languages in some locales and if it'd be affecting the display of some patterns.

I'm not sure worry. My thinking was that having it in the locale's language first would make it clearer to people making use of the theme but I'll check in with a couple of people on translation teams.

juanfra:
Thank you Peter. 

@beafialho as a non-native English speaker and designer of the theme what are your thoughts on this? "Stories" translates to "Histórias" in Portuguese. 

@carolinan do you have any thoughts on this?

Thank you all!

patterns/banner-poster.php
```diff
@@ -21,7 +21,12 @@
 			<!-- wp:column {"width":"80%"} -->
 			<div class="wp-block-column" style="flex-basis:80%">
 				<!-- wp:heading {"textAlign":"left","align":"wide","style":{"typography":{"fontSize":"12vw","lineHeight":"0.9","fontStyle":"normal","fontWeight":"300"}}} -->
-				<h2 class="wp-block-heading alignwide has-text-align-left" style="font-size:12vw;font-style:normal;font-weight:300;line-height:0.9"><?php echo esc_html_x( '“Stories, historias, iсторії, iστορίες”', 'Sample heading in four languages.', 'twentytwentyfive' ); ?></h2>
+				<h2 class="wp-block-heading alignwide has-text-align-left" style="font-size:12vw;font-style:normal;font-weight:300;line-height:0.9">
+					<?php
+						/* translators: This string contains the word "Stories" in four different languages with the first item in the locale's language. */
+						echo esc_html_x( '“Stories, historias, iсторії, iστορίες”', 'Placeholder heading in four languages.', 'twentytwentyfive' );
```
audrasjb: would it be possible to replace esc_html_x with a more general KSES function? Because in the current proposal, we can't add any html tag in this string, so we can't comply with accessibility rules that stipulates that languages changes must be indicated, for example via `<span lang="en">Stories</span>`

Also it would be nice to indicate the lang of each of the four words in the translator's comment

audrasjb (changes requested):
Left a comment about the accessibility of this string :)

audrasjb:
I don't have any issue with keeping this text untranslated in French locale, but if we want this theme to be `accessibility-ready` (which should be our goal), I need to indicate language changes with HTML markup.

beafialho:
> @beafialho as a non-native English speaker and designer of the theme what are your thoughts on this? "Stories" translates to "Histórias" in Portuguese.

@juanfra that's right, "Histórias" is the correct translation for Portuguese.

> Should we give the direction of having the first word in the locale's language or should we leave it open so that the translators take care of it?

Having the first word in the locale's language seems ok to me.

patterns/banner-poster.php
swissspidy: Yes, `lang` attributes would be good. That's also gonna make the string a bit more difficult to grasp for translators. :S

Indicating the language would also be helpful.

juanfra: Thanks for the input!

> Also it would be nice to indicate the lang of each of the four words in the translator's comment

If we go the KSES route and add a `lang` attribute for each, then would it still make sense to have these strings translatable? What would be the reasoning? As a translator, I'd be really confused.

audrasjb:
I'm not opposed to have multilangual strings, but to claim the `accessibility-ready` tag, Twenty Twenty-Five will have to depend on polyglots using correct html markup.
Another solution is to make this string untranslatable, and to hardcode these `<span lang="XX">` tags for each individual string.
Eventually, this would be our best option here.

peterwilsoncc:
> Another solution is to make this string untranslatable, and to hardcode these `<span lang="XX">` tags for each individual string.

I'd like to keep it translatable so the locale's language can be included in the string. If the string was `Historias, iсторії, iστορίες, истории` I'd have no idea what it was conveying so I imagine the same would be true for the roughly 50% of people using a non-english version of WordPress.

I've asked in the [#polyglots](https://wordpress.slack.com/archives/C02RP50LK) slack channel (requires login via [chat.w.org](http://chat.w.org)) for approval to update the string to be accessible as we a re post string freeze. 

I'll push a fix to include `lang` attributes shortly.

audrasjb:
Thanks @peterwilsoncc the new changeset looks good to me! 👍

audrasjb (approved):

peterwilsoncc:
@tobifjellner has signed off on the change in the polyglots slack channel, see [this tread](https://wordpress.slack.com/archives/C02RP50LK/p1729806066405179).

patterns/event-rsvp.php
```diff
@@ -23,7 +23,14 @@
...
 			<div class="wp-block-group">
 				<!-- wp:heading {"fontSize":"xx-large"} -->
-				<h2 class="wp-block-heading has-xx-large-font-size"><?php echo esc_html_x( '“Stories, historias, iсторії, iστορίες”', 'Placeholder heading in four languages.', 'twentytwentyfive' ); ?></h2>
+				<h2 class="wp-block-heading has-xx-large-font-size">
+				<?php
+				echo wp_kses_post(
+					/* translators: This string contains the word "Stories" in four different languages with the first item in the locale's language. */
+					_x( '“Stories, <span lang="es">historias</span>, <span lang="uk">iсторії</span>, <span lang="el">iστορίες</span>”', 'Placeholder heading in four languages.', 'twentytwentyfive' )
+				);
+				?>
```
SergeyBiryukov:
```suggestion
					<?php
					echo wp_kses_post(
						/* translators: This string contains the word "Stories" in four different languages with the first item in the locale's language. */
						_x( '“Stories, <span lang="es">historias</span>, <span lang="uk">iсторії</span>, <span lang="el">iστορίες</span>”', 'Placeholder heading in four languages.', 'twentytwentyfive' )
					);
					?>
```
Should this be indented one level further, like in the other files?

SergeyBiryukov:
Left a minor formatting suggestion, looks good to me otherwise 👍

patterns/banner-poster.php
carolinan: All text needs to be translatable.

carolinan:
The lang attribute is not a requirement for the accessibility-ready tag: https://make.wordpress.org/themes/handbook/review/accessibility/required/

Issue #605, opened 2024-10-21
Dusk: Annotation vertical padding looks uneven


juanfra:
**Description**

The annotation style in the dusk variation looks a bit uneven. It looks like it has more padding-bottom.

**Step-by-step reproduction instructions**

1. Activate dusk.
2. Try out a pattern including the annotation style, or the "Photo blog home"
3. Confirm that the annotation style is looking uneven.

**Screenshots**

![Image](https://github.com/user-attachments/assets/089067d3-becb-4796-b49c-f99bbf35c330)

carolinan:
I can reproduce this, I think it may just be a typo:
I think the bottom padding should just be 0.2rem

```
    padding-top: 0.2rem;
    padding-right: 0.6rem;
    padding-bottom: 0.25rem;
    padding-left: 0.6rem;
```

carolinan:
But it may be better to only update it for this style variation.

carolinan:
Meh, now when I have switched computer, 0.2rem is also too large, and I need to set it to 0 bottom padding for the text to align. 🤷‍♀

carolinan:
Ignore me, it is the same on both the mac (small screen resolution) and PC (large screen resolution), when 0.2 rem is used, the text is still miss-aligned.

troychaplin:
@carolinan setting the padding bottom to 0.2rem seems to be fine, at least to me. Top example is 0.2, while bottom is original of 0.25

![Image](https://github.com/user-attachments/assets/03cb852e-5608-4667-a826-51ce4dad7aa4)

juanfra:
@troychaplin It looks like the problem is with the headings' typography. The paragraph is using Fira sans, and it seems to work fine there. It looks off when the annotation is applied to a heading (using Vollkorn) like the one we have for the home page.

https://github.com/user-attachments/assets/c67c45cd-360a-4bf9-a834-bc2597f44398

troychaplin:
Removing me as assignee in case someone else wants to take a look and has another approach to solving this issue. Notes around what has been attempted to date can be seen in close PR #636

PR #636, opened 2024-10-26
add: spacing styles for heading with annotation style in dust theme

troychaplin:
**Description**

Updated spacing as discussed in #605.

I set the bottom padding to 0.15rem instead of the originally suggested 0.2 based on a heading set in all caps still appearing to have more bottom space than the top.

**Screenshots**

test-2
test-1

### DIFF
--- a/styles/03-dusk.json
+++ b/styles/03-dusk.json
@@ -60,6 +60,7 @@
 			"letterSpacing": "-0.18px",
 			"lineHeight": "1.5"
 		},
+		"css": ".wp-block-heading.is-style-text-annotation {padding-bottom:0.15rem;}",
 		"blocks": {
 			"core/code": {
 				"color": {


rejaulalomkhan:
@troychaplin Thanks for your valuable PR. 
Perfect, thanks for addressing the spacing issue as discussed in #605. The adjustment to 0.15rem for bottom padding makes sense given the heading's appearance. Nice work.

styles/03-dusk.json
```diff
@@ -60,6 +60,7 @@
 			"letterSpacing": "-0.18px",
 			"lineHeight": "1.5"
 		},
+		"css": ".wp-block-heading.is-style-text-annotation {padding-bottom:0.15rem;}",
```
juanfra, line 63: I still see a bit more spacing at the bottom. In my testing I believe approaching `0.1rem` makes it look better. I'll defer to @beafialho to decide the value here.

juanfra (commented):
Thank you, Troy!

This would be initially fixing the annotation padding for the dusk variation. The problem is that once saved, the CSS will be persistent and if you change the style variation then it'll affect how the annotation is viewed. The result would be creating this new problem:

https://github.com/user-attachments/assets/ec652d23-5c36-4b15-8d6d-85f458812ae2

styles/03-dusk.json
troychaplin: Ok, I'll make any additional changes necessary once we hear back

troychaplin: @juanfra stepped away for a but, just following up on a few items. Are we still waiting for confirmation on this PR?

juanfra: @troychaplin thanks for coming back to this one. Unfortunately, as mentioned in the review, the PR will have the side effect of keeping/persisting the CSS once saved, affecting styles when changing variation. We need to find a solution that doesn't have this type of side effects. If we find it, we could confirm the padding-bottom of the annotation style for that variation in particular.

troychaplin: @juanfra it sounds like it might be best to just close this PR and remove my name from the issue. I will still take some time to look at other solutions, but best to leave it unassigned in case someone else wants to take a look as well.

juanfra: Ok, thank you Troy!

troychaplin:
Closing pull request to pave the way for someone else who may have a new or better approach to solving this issue

PR #641, opened 2024-10-28
Adjust outline styles on the navigation block

### LINKED ISSUE #595: Outline styles broken and missing padding
**Description**
If you navigate via keyboard you see the focus styles for the navigation and in the main menu there is missing some padding. Also in submenus the bottom outline is hidden, looks broken.

**Screenshots**
Image
Image

**Environment info**
 - Device: MacBook Pro
 - OS: macOS 12.7.6
 - Browser: Firefox 
 - Version: 131.0.3 (aarch64)

carolinan:
**Description**
Closes https://github.com/WordPress/twentytwentyfive/issues/595

This pull request adjusts the focus outline style on the navigation block to work better with the thicker outline.

1. Increase the bottom margin on submenus, so that the outline is visible.
2. Increase the outline offset on the parent menu items, so that the outline does not touch the text.
3. Then removes that outline offset from the submenus, otherwise the outline is visible outside the submenu container.

I was not able to write 2 and 3 as one style.

**Screenshots**

In this clip I am using  Firefox on Windows 11 and I have zoomed in to 240 % only to make the style more visible for the video:

https://github.com/user-attachments/assets/d16e60cb-2114-45f3-a3b3-24f95deb163f

**Testing Instructions**
Add two navigation blocks, one with the setting "Open on click" disabled and one with it enabled.
Add submenus with at least one nested submenu. (A submenu with a submenu).

Test with different browsers. 
Use the tab key to navigate forwards through the menu.
Use tab key  and backspace to navigate backwards through the menu.

Bonus: Test other blocks that can be added to the menu and make sure that they do not break.

Note: You may see two unrelated issues:
One is a focus loss when navigating backwards.
The other is a styling problem with the nested submenu where the text has no spacing on the left side. These are issue with the block, that are unrelated to this theme.

### DIFF
--- a/style.css
+++ b/style.css
@@ -29,6 +29,21 @@ a {
 	outline-style: solid;
 }
 
+/* Increase the bottom margin on submenus, so that the outline is visible. */
+.wp-block-navigation .wp-block-navigation-submenu .wp-block-navigation-item:not(:last-child) {
+	margin-bottom: 3px;
+}
+
+/* Increase the outline offset on the parent menu items, so that the outline does not touch the text. */
+.wp-block-navigation .wp-block-navigation-item .wp-block-navigation-item__content {
+	outline-offset: 4px;
+}
+
+/* Remove outline offset from the submenus, otherwise the outline is visible outside the submenu container. */
+.wp-block-navigation .wp-block-navigation-item ul.wp-block-navigation__submenu-container .wp-block-navigation-item__content {
+	outline-offset: 0;
+}
+
 /*
  * Progressive enhancement to reduce widows and orphans
  * https://github.com/WordPress/gutenberg/issues/55190


juanfra (commented):
Thanks Carolina. Outline styles are working well to me.

https://github.com/user-attachments/assets/d770d1ec-1f45-44c9-9e4f-1ff3a8cfcffc

Do we want to have the CSS in the `style.css` file or keep it (like we do for other elements) in the `theme.json`?

The other tiny thing I noticed in my testing, unrelated to this PR, is that when having "open to click" enabled, the parent of the third level item (in the screencast "Style variations") looks misaligned.

troychaplin:
@carolinan just tested this and it looks and works nicely!

carolinan:
@juanfra Yes the misalignment of that item is what I meant with this comment:
>The other is a styling problem with the nested submenu where the text has no spacing on the left side. These are issue with the block, that are unrelated to this theme. 

I have not found a matching issue on the Gutenberg GitHub repository yet because... there are too many issues reported for this block 🤐 

> Do we want to have the CSS in the style.css file or keep it (like we do for other elements) in the theme.json?

I think it is too big to add to theme.json?
Another more involved option would be to add it to a separate stylesheet that is only loaded when the navigation block is on the page. But that is not necessarily an improvement since the amount of CSS is small.

juanfra:
> I think it is too big to add to theme.json?
Another more involved option would be to add it to a separate stylesheet that is only loaded when the navigation block is on the page. But that is not necessarily an improvement since the amount of CSS is small.

Ok, if it's too big for the `theme.json` then I think what you pushed is a good option.

juanfra (approved):

elbsegler63:
I tested it as follows:
WP: 6.7-RC1-59297
Browser: Chrome Version 130.0.6723.70
Browser: Firefox 130.0b9
Windows 11

If I tab backwards from the 2nd submenu to the first submenu, the left side of the frame is unfortunately missing.

Screenshot 2024-10-28 183017

carolinan:
Yes thank you @elbsegler63 this was already reported in the issue.

This is because the submenu items overlap. Most likely to ensure that there is no gap between the items when using a mouse to hover over them.
This style that positions the submenus is added by the block, not by the theme.

elbsegler63:
Thank you for the correction. I had overlooked that. 
And many thanks for your efforts.

PR #648, opened 2024-10-28
fix: layout breaks when there are many categories in default single p…

### LINKED ISSUE #644: Default single post template: "Post meta" Layout breaks when there are many categories
**Description**

By chance I was testing a Gutenberg PR with the theme test data installed.
In the theme test data there is a post with many categories. This is an edge case, but in this post,
the number of categories breaks the layout of the row block that displays the author name and categories below the featured image.

Install https://github.com/WordPress/theme-test-data
View the Edge Case: Many Categories post.

**Screenshots**

Image

**Additional context**

By enabling the option "Allow to wrap to multiple lines" on the parent row block, the design could be preserved like this:
Image

alaminfirdows:
**Description**
Closes #644

This PR fixes the issue where the layout breaks when there are many categories in the default single post template. Changing the `flexWrap` property of the parent group from `nowrap` to `wrap` resolves this.

**Screenshots**
image

**Testing Instructions**

Install https://github.com/WordPress/theme-test-data
view the "Edge Case: Many Categories" post.

Alternatively, create a post with more categories (about 10+) and view the post.

### DIFF
--- a/patterns/hidden-written-by.php
+++ b/patterns/hidden-written-by.php
@@ -10,7 +10,7 @@
  */
 
 ?>
-<!-- wp:group {"style":{"spacing":{"blockGap":"0.2em","margin":{"bottom":"var:preset|spacing|60"}}},"textColor":"accent-4","fontSize":"small","layout":{"type":"flex","flexWrap":"nowrap"}} -->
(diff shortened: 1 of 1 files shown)


juanfra (approved):
Thank you! Looks good to me.

https://github.com/user-attachments/assets/e6679673-362b-4ce0-bb90-768b2b9c2020

PR #650, opened 2024-10-28
Twilight, Section 5: Adjust the border color for search input

### LINKED ISSUE #649: Twilight, Section 5: Input field of the core/search block are not visible
**Description**
Related #640

In the Twilight theme variation, section style 5 has a white background. The accent color `#FFFFFF33`, which is used as a border for the core/search block's input field, doesn’t appear clearly. Since the background of the elements is also white, the input fields are difficult to see.

1. Navigate to **Appearance > Editor > Styles** in your WordPress dashboard.
2. Select and enable the **Twilight** style variation.
3. Add a **Group** block, and apply **Section Style 5** to it.
4. Inside this group, add a **Search** block.
5. Finally, view the section that includes the comments.

**Expected behavior**
Users should be able to select and use the form field easily.

**Screenshots**
Image

alaminfirdows:
**Description**
Closes #649

Adjust the border color on form elements in the search form block to make the inputs stand out against the white background. Previously, they were hard to see because the accent color was too similar to the background color.

**Screenshots**
image

**Testing Instructions**
1. Open the **Appearance** menu, then go to **Editor** > **Styles**.
2. Turn on the **Twilight** style variation.
3. Create a new page.
4. Add a **Group Block** to the page, then select **Section Style 5** for it.
5. Inside this section, insert a **Search Block**.
6. Visit the page to see the section with the search block displayed.

### DIFF
--- a/styles/05-twilight.json
+++ b/styles/05-twilight.json
@@ -245,7 +245,7 @@
 						"css": "& textarea, input:not([type=submit]){border-radius:.25rem; border-color: color-mix(in srgb, currentColor 20%, transparent) !important;} & input[type=checkbox]{margin:0 .2rem 0 0 !important;} & label {font-size: var(--wp--preset--font-size--small); }"
 					},
 					"core/search": {
-						"css": "& .wp-block-search__input{border-color: color-mix(in srgb, currentColor 20%, transparent) !important;}"
+						"css": "& .wp-block-search__input{border-color: color-mix(in srgb, currentColor 20%, transparent);}"
 					}
 				}
 			}
--- a/styles/05-twilight.json
+++ b/styles/05-twilight.json
@@ -243,6 +243,9 @@
 				"blocks": {
 					"core/post-comments-form": {
 						"css": "& textarea, input:not([type=submit]){border-radius:.25rem; border-color: color-mix(in srgb, currentColor 20%, transparent) !important;} & input[type=checkbox]{margin:0 .2rem 0 0 !important;} & label {font-size: var(--wp--preset--font-size--small); }"
+					},
+					"core/search": {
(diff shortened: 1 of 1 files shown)


styles/05-twilight.json
```diff
@@ -243,6 +243,9 @@
 				"blocks": {
 					"core/post-comments-form": {
 						"css": "& textarea, input:not([type=submit]){border-radius:.25rem; border-color: color-mix(in srgb, currentColor 20%, transparent) !important;} & input[type=checkbox]{margin:0 .2rem 0 0 !important;} & label {font-size: var(--wp--preset--font-size--small); }"
+					},
+					"core/search": {
+						"css": "& .wp-block-search__input{border-color: color-mix(in srgb, currentColor 20%, transparent) !important;}"
```
juanfra:
```suggestion
						"css": "& .wp-block-search__input{border-color: color-mix(in srgb, currentColor 20%, transparent);}"
```

juanfra (commented):
Thank you! Works well. I believe the `!important` is not needed here.

alaminfirdows:
Thanks @juanfra.

I've also noticed that there is some additional override code for the comment block. Shouldn't we consider removing it to avoid duplicating the code? 

Ref: https://github.com/WordPress/twentytwentyfive/pull/650/files/a30575ef27ca892f079b2606a5120d2976d229d3#diff-60a85af5c000152e40c3680e65867153567c39f66b221f8200e757de4d123f5bR245

juanfra:
I am not sure if I understand. The two overrides are needed to fix the bug with the comments block and the search block. Can you please elaborate on why we should remove it?

alaminfirdows:
I wanted to point out that the comment block override has some additional changes that we don't really need. We should keep just the border color, but we can remove the other properties, such as border-radius, margin, and font-size. 

What do you think about this?

PR #656, opened 2024-10-29
Remove text color from the term description block

### LINKED ISSUE #655: Term description has too low color contrast on some style variations
**Description**

In theme.json, the term description block is set to use accent color four. 
The block is not styled anywhere else in any style variation or preset.

Accent color four has too low contrast when it is used for the text color on top of the background color in some section styles:
1. Default, section style 4 and 5
2. Midnight, section style 1 and 5
3. Noon, section style 4 and 5
4. Dusk, section 2, 4, 5

I stopped testing here. Note that this block can have both plain text and links. I did not test the link styles.

Dusk:
Image

The term description block is easiest to test on archive templates, since this type of template is the only template that shows the block on the front.

Go to Appearance > Editor > Templates
Select to edit the archive template.
This template already has a term description block placed, but to test the colors, you need to place it inside a group (or column) with
a section style applied.
- Then change the section style and the theme style variation to test the possible combinations.

**Additional context**

Instead of solving this in each variation and section style, my recommendation would be to remove the text color from this block in theme.json, and default to whatever text color is set in the variations.

carolinan:
**Description**
By removing the text color from the term description block, the number of color contrast issues when the block is used in a "section" is reduced.

Closes https://github.com/WordPress/twentytwentyfive/issues/655

**Screenshots**
Because of the large number of contrast issues, I will not provide screenshots or videos of all before and after scenarios.
***After***

https://github.com/user-attachments/assets/be6081ac-310d-4ff1-a637-089aa40980f3

**Testing Instructions**
The term description block is easiest to test on archive templates, since this type of template is the only template that shows the block on the front.

Go to Appearance > Editor > Templates
Select to edit the archive template.
This template already has a term description block placed, but to test the colors, you need to place it inside a group (or column) with a section style applied.

Then change the section style and the theme style variation to test the possible combinations.

### DIFF
--- a/theme.json
+++ b/theme.json
@@ -568,9 +568,6 @@
 				}
 			},
 			"core/term-description": {
-				"color":{
-					"text": "var:preset|color|accent-4"
-				},
 				"typography": {
 					"fontSize": "var:preset|font-size|medium"
 				}


juanfra (approved):
Thanks! Looks good to me

https://github.com/user-attachments/assets/9f251ea8-3fe8-46c4-a572-891a99696912

Issue #664, opened 2024-11-04
Full page patterns have redundant nested groups which makes section styles unpredictable


jasmussen:
**Description**

The default full page patterns have multiple nested redundant groups. This conflicts with zoom out mode and section styles, which is best described first visually in this GIF:

![Image](https://github.com/user-attachments/assets/2a13bb7e-9a21-4a4b-ba2f-69a172cc262d)

**Step-by-step reproduction instructions**

1. Go to site editor.
2. Create a NEW page, give it any title.
3. Choose the "Business homepage" pattern.
4. Click a section to change the section style to a different color.
5. Now zoom out, observe how a different section style is chosen, and choosing a new section style isn't actually applied.

What happens here is that zoom out mode works with top level sections, allowing you to select only those. Which means if there's a wrapper group that is _not_ the source of the section style, then there will be a perceived invisible disconnect, as the section style you just applied is one level deeper, on a block no longer selected.

**Expected behavior**

If I set a section style not zoomed out, then zoom out, I should see the same section style. Ideally, we can remove the redundant wrapper groups. Observe in this screenshot, for example, how the "fleurs co" pattern has two wrapping groups:

 ![Image](https://github.com/user-attachments/assets/ae27f915-d46c-4456-acc7-08c2656b52f2)

Because of the two groups, when you click not zoomed in, you'll select the most nested one. But if you click when zoomed out, you'll select the _least_ nested one. 

**Environment info**

6.7-rc2

jasmussen:
@juanfra @carolinan @beafialho do you know why there are additional wrapping groups in these patterns? And would they be safe to remove?

juanfra:
@jasmussen the wrapping groups were added [here](https://github.com/WordPress/twentytwentyfive/pull/514), and it was in combination with some other fixes, trying to resolve everything that was reported [here](https://github.com/WordPress/twentytwentyfive/issues/497). 

These wrapping groups were added because the landing pages were built (originally) using one group that wrapped all inner patterns, and that was causing interferences with the zoom-out view. The other reason is that we needed to set a zero margin between the groups so that the layout respected the design (no vertical gap between full-width groups). If we put one pattern after the other, we would have a gap between them.

However, we can try to see how removing the groups from these pages reacts to the most recent improvements to the zoom-out functionality.

jasmussen:
[...]

PR #666, opened 2024-11-05
Fix group block width for news blog home.

### LINKED ISSUE #665: Expected patterns to be fullwidth so that section styles work
**Description**
I expected these patterns used would be full width, so that one could apply section styles to them without the patterns looking broken. Reviewing based on [this Playground link](https://playground.wordpress.net/?blueprint-url=https%3A%2F%2Fraw.githubusercontent.com%2FWordPress%2Fblueprints%2Ftt5demo%2Fblueprints%2Ftt5-demo%2Fblueprint.json). 

**Screenshots**

Image

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/665

**Screenshots**

https://github.com/user-attachments/assets/81e9d70c-2d96-4479-aaf6-a41b11686e43

**Testing Instructions**

1. Go to the site editor.
2. Edit the home page.
3. Select the "News blog home" design.
4. Confirm that the main groups are full width, and that applying the section styles to them doesn't look broken.

### DIFF
--- a/patterns/template-home-news-blog.php
+++ b/patterns/template-home-news-blog.php
@@ -14,7 +14,7 @@
 ?>
 <!-- wp:template-part {"slug":"header-large-title"} /-->
 
-<!-- wp:group {"tagName":"main","layout":{"type":"constrained"}} -->
+<!-- wp:group {"tagName":"main","style":{"spacing":{"blockGap":"0"}},"layout":{"type":"constrained"}} -->
 <main class="wp-block-group">
 	<!-- wp:group {"align":"full","style":{"spacing":{"padding":{"top":"var:preset|spacing|50","bottom":"var:preset|spacing|50"}}},"layout":{"type":"constrained"}} -->
 	<div class="wp-block-group alignfull" style="padding-top:var(--wp--preset--spacing--50);padding-bottom:var(--wp--preset--spacing--50)">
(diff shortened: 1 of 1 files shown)


jasmussen (approved):
Confirmed, this improves it, thank you:

image

Should we apply block-spacing 0 to the wrapping Group block, to zero out the margin between colored sections? Zoom out can "punch through" the top level singular group wrapper and select the items inside. In fact this might also be a solution for #664. See screenshot above.

juanfra:
> Should we apply block-spacing 0 to the wrapping Group block, to zero out the margin between colored sections? 

Sounds good, done ✅

> Zoom out can "punch through" the top level singular group wrapper and select the items inside. In fact this might also be a solution for https://github.com/WordPress/twentytwentyfive/issues/664. See screenshot above.

It seems that in this context, it can. But it's not behaving like that in pages or posts; that was one of the reasons we ended up having multiple wrapping groups.

jasmussen:
> It seems that in this context, it can. But it's not behaving like that in pages or posts; that was one of the reasons we ended up having multiple wrapping groups.

It seems like this might thread the needle though: allow us to use includes to import the patterns, just wrap it in a single group that sets the block gap. In doing so, we won't need the redundant groups, and zoom out will still select the correct "top level" section for applying section styles. What do you think?

juanfra:
@jasmussen Ok, so I checked and to get this behavior, the single group should have the `<main>` tag. 

I definitely prefer to have the landing pages wrapped in a `<main>` than duplicating patterns. This would be the best option we have at this moment, in my opinion.

I'll open a PR with those changes for you to check out.

jasmussen:
Thank you! And this PR seems good to go too!

PR #667, opened 2024-11-05
Try: Reduce the number of groups in landing pages to improve zoom out experience.

### LINKED ISSUE #664: Full page patterns have redundant nested groups which makes section styles unpredictable

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/664

We're using a `<main>` wrapper for the landing pages with a block gap of zero, so that the different sections are editable in zoom-out view.

**Screenshots**

https://github.com/user-attachments/assets/33da1deb-bdb3-4442-af64-9d58d6a2c91b

**Testing Instructions**

1. Create a page.
2. Insert a landing page. Choose from: "Shop homepage", "Business homepage", "Landing page for book", "Landing page for event", "Landing page for podcast".
3. Confirm that the design is ok.
4. Confirm that the styles of the main sections of the landing pages are accessible in zoom out view.

### DIFF
--- a/patterns/grid-with-categories.php
+++ b/patterns/grid-with-categories.php
@@ -26,8 +26,11 @@
 		<!-- wp:group {"layout":{"type":"flex","orientation":"vertical","justifyContent":"stretch"}} -->
 		<div class="wp-block-group">
 			<!-- wp:cover {"url":"<?php echo esc_url( get_template_directory_uri() ); ?>/assets/images/category-anthuriums.webp","id":2904,"alt":"Close up of a red anthurium.","dimRatio":0,"customOverlayColor":"#833d3a","isUserOverlayColor":true,"layout":{"type":"constrained"}} -->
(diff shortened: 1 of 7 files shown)


jasmussen:
Gave this a quick test in the site editor > pages > new flow, and I'm seeing a different issue there, where only the business homepage can be selected:

2024-11-05 12 53 52 tt5 local 92ff6d758240

I'm also still seeing the redundant groups here:

Screenshot 2024-11-05 at 12 54 35

Did I test things wrong, here? 

Thanks for the PR!

juanfra:
I see. In that flow, the content is already the "main", so we would have two "main" and the fix won't work.

> I'm also still seeing the redundant groups here:

Those are coming from the pattern.

Edit: In the flow of creating a page and inserting the landings, we'll also have the markup with two `<main>` elements and that creates accessibility problems.

jasmussen:
Any instincts, @richtabor ?

juanfra:
Closing in favor of https://github.com/WordPress/twentytwentyfive/pull/669

I also cherry-picked the other fixes and created a PR [here](https://github.com/WordPress/twentytwentyfive/pull/668).

PR #669, opened 2024-11-05
Try: Remove wrappers from landing pages, and set zero margin to individual patterns.

### LINKED ISSUE #664: Full page patterns have redundant nested groups which makes section styles unpredictable

juanfra:
**Description**

Fixes https://github.com/WordPress/twentytwentyfive/issues/664

We're removing the wrappers of the landing pages, so that they're functional in the zoom out view. To respect the designs, we are also setting a zero margin to the patterns that are involved in these landing pages.

**Screenshots**

https://github.com/user-attachments/assets/9dde23f0-0844-4e21-a565-ab01c0e4d8e8

**Testing Instructions**

1. Create a page.
2. Insert a landing page. Choose from: "Shop homepage", "Business homepage", "Landing page for book", "Landing page for event", "Landing page for podcast".
3. Confirm that the design is ok.
4. Confirm that the styles of the main sections of the landing pages are accessible in zoom out view.

Repeat again, creating a page from the site editor.


jasmussen (approved):
Thanks for this, this works well for me:







It simplifies all the patterns to have a more basic nesting structure, and in the process it also ensures that selecting a section and applying a section style works the same, and as expected, whether you're zoomed or not!

richtabor:
What do you think about applying margin 0 for any fullwidth patterns that can be inserted? That way they all work the same, and you don't end up with this: 

CleanShot 2024-11-05 at 09 39 56

jasmussen:
> What do you think about applying margin 0 for any fullwidth patterns that can be inserted? That way they all work the same, and you don't end up with this:

I'd support that. We'd still need the fewer wrappers, but it could simplify the code, no?

juanfra:
> What do you think about applying margin 0 for any fullwidth patterns that can be inserted? That way they all work the same, and you don't end up with this:

@richtabor @jasmussen the problem with setting all of them to zero is that the layout setting for block spacing won't affect any pattern.

jasmussen:
> the problem with setting all of them to zero is that the layout setting for block spacing won't affect any pattern.

I think that's fine, it will still affect paragraphs and other bits and pieces. I don't think there's an expectation that this particular property would affect the sections as such.

juanfra:
Ok, I updated the rest of the full width patterns.

jasmussen:
Ship it!

